PR #1, opened 2018-05-05
Project structure for review.

fexolm:


src/main.go
```diff
@@ -0,0 +1,45 @@
...
+var checkers = []checker{}
+
+func main() {
+	fset := token.NewFileSet()
+	dir := flag.String("dir", "", "project directory")
+	flag.Parse()
+
+	if *dir == "" {
+		flag.Usage()
+		os.Exit(1)
```
Quasilyte:
```
		flag.Usage()
		os.Exit(1)
```
It would be helpful for user to get actual error message in this case. Like "illegal empty -dir argument".
You can factor these two lines into function with additional "message" argument.
```go
func blame(format string, args ...interface{}) {
	log.Printf(format, args...)
	flag.Usage()
	os.Exit(1)
}
```
And call it there like:
```go
blame("illegal empty -dir argument\n")
```
The `...args` would be useful, if some of `blame` calls would need formatted output.

src/main.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+type consoleLogger struct{}
+
+func (c consoleLogger) log(format string, args ...interface{}) error {
+	_, err := fmt.Printf(format, args...)
+	return err
+}
+
+var checkers = []checker{}
```
Quasilyte: It may be better to have `checkers` as local variable inside `main`.
Or is there any particular reason for them to be global?

src/main.go
```diff
@@ -0,0 +1,51 @@
...
+	"log"
+	"os"
+)
+
+type logger interface {
+	log(format string, args ...interface{}) error
+}
+
+type checker interface {
+	run(fset *token.FileSet, logger logger) error
```
Quasilyte: We can introduce `lint.Context` type that will hold both `fset`, and `logger`, and perhaps even some (or all) command line flags that we will almost certainly need inside some checkers.
(This means that we need `src/lint/lint.go` file that defines `Context` type; you may call that `Config`, if you find it's more fitting.)

Quasilyte:
This `main.go` file should go to `src/cmd/kfulint/main.go`.

src/main.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+type consoleLogger struct{}
```
Quasilyte: I almost 99% certain that we will only need console (stdout/stderr) loggers. :)

src/main.go
```diff
@@ -0,0 +1,51 @@
...
+
+func blame(format string, args ...interface{}) {
+	log.Printf(format, args...)
+	flag.Usage()
+	os.Exit(1)
+}
+
+func main() {
+	fset := token.NewFileSet()
+	dir := flag.String("dir", "", "project directory")
```
Quasilyte: `s/project directory/package directory/` I do think that we expect single package inside one directory that is being checked. That's the most common case.

src/main.go
```diff
@@ -0,0 +1,51 @@
...
+	flag.Parse()
+
+	if *dir == "" {
+		blame("illegal empty -dir argument\n")
+	}
+
+	parser.ParseDir(fset, *dir, nil, parser.ParseComments)
```
Quasilyte: Maybe we need to grab `ParseDir` result to `lint.Context`.

src/main.go
fexolm: It was for testing purposes. I think it would be nice to have opportunity to write fake logger to write unit tests.

src/main.go
fexolm: I also thank about warning type that will let us write logs in different formats.

src/main.go
fexolm: I also thought about warning model which would allow writing in different formats (like json, xml, etc.)

Quasilyte: We may come up to some warning representation eventually.
What about using `error` for now?
When we will implement several checkers, it will be more apparent what we need and what we want from that type.

Quasilyte:
I see where we can move from here.
It could be better to improve from here in a separate commits and merge this one in it's current state.
(Also want to try "optimistic merging", if you don't mind.)

PR #3, opened 2018-05-05
Fix package url.

fexolm:


Quasilyte:
I will remove package path fix from my PR then.
Merge this, and then we can proceed.

PR #7, opened 2018-05-06
lint: add param-name checker

Quasilyte:
Also modify kfulint/main.go and lint.go to make
it possible to implement first checker.

### DIFF
--- a/lint/param-name_checker.go
+++ b/lint/param-name_checker.go
@@ -1,0 +1,87 @@
+package lint
+
+import (
+	"fmt"
+	"go/ast"
+	"strings"
+)
+
+// ParamNameChecker detects potential issues in function parameter names.
+//
+// Rationale: makes godoc output better.
+type ParamNameChecker struct {
+	ctx *Context
+
+	loudNames map[string]bool
+
+	warnings []Warning
+}
+
+// NewParamNameChecker returns initialized checker for Go functions param names.
+func NewParamNameChecker(ctx *Context) *ParamNameChecker {
+	return &ParamNameChecker{
+		ctx: ctx,
+		loudNames: map[string]bool{
+			"IN":    true,
+			"OUT":   true,
+			"INOUT": true,
+		},
+	}
+}
+
+// Check runs parameter names checks for f.
+//
+// Features
+//
+// 1. Detects somewhat common "loud" identifiers, like IN or OUT.
+//    Suggests to replace them with down-case versions (in, out).
+//
+// 2. If capitalized name is not recognized as "loud",
+//    treat it as "redundantly exported".
+//    Suggests to use non-capitalized identifier.
+func (c *ParamNameChecker) Check(f *ast.File) []Warning {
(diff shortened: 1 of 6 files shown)


lint/param-name_checker.go
```diff
@@ -0,0 +1,74 @@
...
+
+	warnings []Warning
+}
+
+// NewParamNameChecker returns initialized checker for Go functions param names.
+func NewParamNameChecker(ctx *Context) *ParamNameChecker {
+	return &ParamNameChecker{
+		ctx: ctx,
+		loudNames: map[string]bool{
+			"IN":    true,
```
fexolm: Is there any reason to use map instead on set here?

lint/param-name_checker.go
```diff
@@ -0,0 +1,74 @@
...
+//    treat it as "redundantly exported".
+//    Suggests to use non-capitalized identifier.
+func (c *ParamNameChecker) Check(f *ast.File) []Warning {
+	c.warnings = c.warnings[:0]
+	for _, decl := range collectFuncDecls(f) {
+		for _, param := range collectFuncParams(decl) {
+			for _, id := range param.Names {
+				switch {
+				case c.loudNames[id.Name]:
+					c.warnLoud(id)
```
fexolm: Maybe we need also add camel case check here?

lint/util.go
```diff
@@ -0,0 +1,32 @@
...
+	for _, decl := range f.Decls {
+		if decl, ok := decl.(*ast.FuncDecl); ok {
+			decls = append(decls, decl)
+		}
+	}
+	return decls
+}
+
+func collectFuncParams(decl *ast.FuncDecl) []*ast.Field {
+	var params []*ast.Field
```
fexolm: It seems line that this function will work only in param_name_checker. Do we really need array with recv, params and results in global case? Maybe we should move this function in param_name_checker or slpit it into 3 functions?

lint/param-name_checker.go
Quasilyte: `golint` does that.

lint/param-name_checker.go
Quasilyte: There are 2 idioms for "sets" in Go:
- `map[T]bool`
- `map[T]struct{}`
The latter was more efficient in the past, not sure it's accurate anymore.
The fist one is preferred, you can check if element is in set by simply doing `m[key]`.

No "real" sets are builtin or inside stdlib.

lint/util.go
Quasilyte: Good point.
Fixed.

Quasilyte:
We could force checkers to return `([]Warning, error)` and avoid panic for "exceptional situations", but:
1. I think most checkers will always return `nil` error.
2. Even if they want to return some kind of fatal error, it could be inconvenient to do so due to complications with `ast.Walk` and functions alike (can't just return from containing function while traversing AST with callback function).

PR #19, opened 2018-05-08
Quasilyte/type guard v1

Quasilyte:
Does trigger for cases from `testdata/type-guard`.
Does not handle nested type switches if containing type switch already triggered warning.
Will fix that later (todo is on it's place).

### DIFF
--- a/lint/type-guard_checker.go
+++ b/lint/type-guard_checker.go
@@ -36,7 +36,7 @@ func (c *TypeGuardChecker) Check(f *ast.File) []Warning {
 				c.check(stmt)
 				return false
 			}
-			return false
+			return true
 		})
 	}
 	return c.warnings
@@ -78,7 +78,7 @@ func (c *TypeGuardChecker) findTypeAssert(root ast.Stmt, expr, typ ast.Expr) boo
 				astcmp.EqualExpr(typ, assert.Type)
 			return !found
 		}
-		return false
+		return true
 	})
 	return found
 }
--- a/lint/type-guard_checker.go
+++ b/lint/type-guard_checker.go
@@ -28,12 +28,15 @@ func (c *TypeGuardChecker) Check(f *ast.File) []Warning {
 	// may have other type switches inside their clauses that
 	// can trigger warnings as well.
 	for _, decl := range collectFuncDecls(f) {
+		if decl.Body == nil {
+			continue
+		}
 		ast.Inspect(decl.Body, func(x ast.Node) bool {
 			if stmt, ok := x.(*ast.TypeSwitchStmt); ok {
 				c.check(stmt)
 				return false
 			}
-			return x != nil
+			return false
 		})
 	}
 	return c.warnings
@@ -75,7 +78,7 @@ func (c *TypeGuardChecker) findTypeAssert(root ast.Stmt, expr, typ ast.Expr) boo
 				astcmp.EqualExpr(typ, assert.Type)
 			return !found
 		}
-		return x != nil
+		return false
 	})
 	return found
 }
--- a/lint/type-guard_checker.go
+++ b/lint/type-guard_checker.go
@@ -73,10 +73,10 @@ func (c *TypeGuardChecker) checkTypeSwitch(root *ast.TypeSwitchStmt) {
 	}
 }
 
-func (c *TypeGuardChecker) warn(stmt *ast.TypeSwitchStmt, caseIndex int) {
+func (c *TypeGuardChecker) warn(node ast.Node, caseIndex int) {
 	s := "case %d can benefit from type switch with assignment"
 	c.warnings = append(c.warnings, Warning{
-		Node: stmt,
+		Node: node,
 		Text: fmt.Sprintf(s, caseIndex),
 	})
 }
--- a/lint/type-guard_checker.go
+++ b/lint/type-guard_checker.go
@@ -24,44 +24,44 @@ func NewTypeGuardChecker(ctx *Context) *TypeGuardChecker {
 // Check runs type switch checks for f.
 func (c *TypeGuardChecker) Check(f *ast.File) []Warning {
 	c.warnings = c.warnings[:0]
-	// TODO(quasilyte): do recursive analysis. Type switches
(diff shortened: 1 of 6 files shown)


Quasilyte:
At least we saw that CI still works. ;)

Quasilyte:
Refs https://github.com/PieselBois/kfulint/issues/10.

lint/type-guard_checker.go
```diff
@@ -18,6 +23,67 @@ func NewTypeGuardChecker(ctx *Context) *TypeGuardChecker {
...
+	// TODO(quasilyte): do recursive analysis. Type switches
+	// may have other type switches inside their clauses that
+	// can trigger warnings as well.
+	for _, decl := range collectFuncDecls(f) {
+		ast.Inspect(decl.Body, func(x ast.Node) bool {
+			if stmt, ok := x.(*ast.TypeSwitchStmt); ok {
+				c.check(stmt)
+				return false
+			}
+			return x != nil
```
fexolm: It doesn't really necessary. Inspect will never follow to nil child.
```
If f returns true, Inspect invokes f recursively 
for each of the non-nil children of node,
followed by a call of f(nil).
```
Also we might neet to check decl.Body for nil befor calling Inspect. Bc inspect shouldn't be invoked with nil argument.
``` Body *BlockStmt    // function body; or nil for external (non-Go) function```

File: lint/type-guard_checker.go [hunk 2]
```diff
@@ -18,6 +23,67 @@ func NewTypeGuardChecker(ctx *Context) *TypeGuardChecker {
...
+		})
+	}
+	return c.warnings
```
fexolm: Warnings seems to never filled

lint/type-guard_checker.go
```diff
@@ -18,6 +23,67 @@ func NewTypeGuardChecker(ctx *Context) *TypeGuardChecker {
...
+func (c *TypeGuardChecker) check(root *ast.TypeSwitchStmt) {
+	if _, ok := root.Assign.(*ast.ExprStmt); !ok {
+		return
+	}
+	assert, ok := root.Assign.(*ast.ExprStmt).X.(*ast.TypeAssertExpr)
+	if !ok {
+		return
+	}
+
+	for i, clause := range root.Body.List {
```
fexolm: We might need to check Body for nil

lint/type-guard_checker.go
Quasilyte: We need to replace `ast.Inspect` because it doesn't terminate traversal even after `false` is returned. But you're right. `x != nil` is not needed. Will fix that.

Good point about `nil` function bodies, thanks.

lint/type-guard_checker.go
Quasilyte: See `func (c *TypeGuardChecker) warn(stmt *ast.TypeSwitchStmt, caseIndex int) {`.
I'm following convention established in `param-name` checker.
`warn` prefixed methods append specific warning.
If kind is not empty, it's appended to method name, like `warnLoud` in case `Kind="Loud"`.

lint/type-guard_checker.go
Quasilyte: See `func (c *TypeGuardChecker) warn(stmt *ast.TypeSwitchStmt, caseIndex int) {`.
I'm following convention established in `param-name` checker.
`warn` prefixed methods append specific warning.
If kind is not empty, it's appended to method name, like `warnLoud` in case `Kind="Loud"`.

lint/type-guard_checker.go
Quasilyte: Don't think it's required.
`ast.TypeSwitchStmt.Body` seems to be non-nil for all cases.
```go
	switch v.(type) {
	case int:
	case float32:
	}
	switch v.(type) {
	}
```
Both type switches pass without panic.

Quasilyte: I will add additional test just in case.

lint/type-guard_checker.go
```diff
@@ -24,63 +24,95 @@ func NewTypeGuardChecker(ctx *Context) *TypeGuardChecker {
...
-		if assert, ok := x.(*ast.TypeAssertExpr); ok {
-			found = astcmp.EqualExpr(expr, assert.X) &&
-				astcmp.EqualExpr(typ, assert.Type)
-			return !found
+// findEqual walks tree root trying to find node that is equal to x.
+// Matched node is returned.
+func (c *TypeGuardChecker) findEqual(root, x ast.Node) (found ast.Node) {
+	defer func() {
+		if r := recover(); r != nil {
+			panic(r)
```
fexolm: Maybe we could write our own utils to walk through ast and not to use panic? Maybe you should write todo here?

Quasilyte: > Maybe we could write our own utils to walk through ast

Maybe we could, but that's inefficient time investment right now.
It will require a lot of tests and API design to be usable.
It's simpler to do simple wrappers around `ast` for now, and
think about better design later, when we have more code to consider our needs.

Working with AST is a pain sometimes, but it can't be solved easily.
There are inherent difficulties in it.

> Maybe you should write todo here?

SGTM.

Quasilyte: There are multiple ways, not even counting making our own traverse functions:
1. Take `golang.org/x/tools/go/ast/astutil` and use it's `Apply`. It's more powerful and has features we want. (Side note, look [how it handles traversal abort](https://github.com/golang/tools/blob/48418e5732e1b1e2a10207c8007a5f959e422f20/go/ast/astutil/rewrite.go#L45-L50)).
2. Use `ast.Walk` with special-purpose visitors that keep `cancel` field inside them and set it to true when certain condition is met.

I'm going to vendor astutil for now.

Quasilyte:
@fexolm, I think AST filters and combinators are a way to go, far better than hardcoded `walkFuncDeclBodies` things.
Also re-implemented `find` function in terms of `astutil.Apply`. Requires no panic now.

fexolm:
@Quasilyte maybe we should merge that pr at last? It becomes large.

Quasilyte:
@fexolm, here goes.

Issue #9, opened 2018-05-07
lint: add unnecessary brackets checker


fexolm:
We need to add checker which will find the errors described below.
 - [x] `var _ [n](*int)` ===>> `var _ [n]*int`
 - [x] `*(name).field` ===>> `name.field` (moved to #14.)
 - [x] `var _ (func())` ===>> `var _ func()`
 - [x] `func(...) (...)` ===>> `func(...) ...`
 - [x] `(*a)[i]` where a is an array (moved to #14.)

Quasilyte:
You can start with a simplest case: fix function result types.
This way, you can concentrate on parenthesis elimination instead of type expression seeking.
Then you can expand the scope to more type expressions.

Quasilyte:
There is also `(*a)[i]` where `a` is an array (not a slice).

Quasilyte:
Do we need [astutil](https://godoc.org/golang.org/x/tools/go/ast/astutil)?
https://github.com/golang/tools/blob/master/go/ast/astutil/util.go#L6

fexolm:
Function return statement case is done in #22.

fexolm:
Checks 2 and 5 moved to #14.

PR #22, opened 2018-05-08
lint: add simple parenthesis check

fexolm:


fexolm:
refs #9

lint/parenthesis-checker.go
```diff
@@ -0,0 +1,46 @@
+package lint
+
+import (
+	"fmt"
+	"go/ast"
+)
+
+// ParenthesisChecker ...
```
Quasilyte: Don't be too lazy.
Either write at least proper sentence, or drop it completely and let linter rage continue (someone will fix that later).
You're only making it harder to refactor code later (fooling around linter).

testdata/foo.go
```diff
@@ -11,3 +11,7 @@ func paramNamesCapitalized0(X, Y int) {}
 func paramNamesCapitalized1(X int) (Y, Z int) { return 0, 0 }
 
 func (PN paramNames) Capitalized2() {}
+
+func badReturn() [](func()) {
```
Quasilyte: You can also add more complex tests like:
```go
func f() [](func([](func()))) {
	return nil
}
```
I think current implementation will only find outermost parenthesis and will not suggest optimal `[]func([]func())` form.

If it's this way, either add a TODO or make a note in referenced issue.
We can improve on this.

lint/parenthesis-checker.go
```diff
@@ -0,0 +1,46 @@
...
+			continue
+		}
+		for _, res := range decl.Type.Results.List {
+			c.validateResultDecl(res)
+		}
+	}
+	return c.warnings
+}
+
+func (c *ParenthesisChecker) validateResultDecl(f *ast.Field) {
```
Quasilyte: We could work on this method design later to make it possible to re-use it for other AST nodes check (we don't want to be limited to results).

But for now, let's change `validate` to `check` for consistency (`checkResultDecl`).

lint/parenthesis-checker.go
Quasilyte: It's actually very common bad practice to use `// Foo ...` or something like that.
We could add a checker for that.

Quasilyte:
And by the way, good job. Made it ahead of other members. :p

lint/parenthesis-checker.go
fexolm: thank you, i have created an issue for that #24

testdata/foo.go
fexolm: You are right, i had added TODO for that.

lint/parenthesis-checker.go
```diff
@@ -5,19 +5,26 @@ import (
...
 
-// ParenthesisChecker ...
+// ParenthesisChecker detects some cases where parenthesis are unnecessary
 type ParenthesisChecker struct {
 	ctx *Context
 
 	warnings []Warning
 }
 
+// NewParenthesisChecker returns initialized checker for Go functions param names.
```
Quasilyte:
```diff
- NewParenthesisChecker returns initialized checker for Go functions param names.
+ NewParenthesisChecker returns initialized checker for type expressions.
```

Quasilyte: I think it will check more than function params eventually.

PR #26, opened 2018-05-08
testdata: move tests to separate dirs

Quasilyte:
The structure will be: `testdata/${checker-name}/checker_tests.go`.
The package name for each test suite is `checher_test`. It doesn't matter right now though.

Discussion is possible. :)


fexolm:
LGTM

PR #28, opened 2018-05-08
all: add more github status checks

weeellz:


lint/util.go
```diff
@@ -19,6 +20,8 @@ func collectFuncDecls(f *ast.File) []*ast.FuncDecl {
 
 func nodeString(fset *token.FileSet, x ast.Node) string {
 	var buf bytes.Buffer
-	printer.Fprint(&buf, fset, x)
+	if err := printer.Fprint(&buf, fset, x); err == nil {
+		log.Fatal(err)
```
Quasilyte: This would normally "never happen", but if it ever-ever does, we need stacktrace for it.
The simplest way for now to achieve that is to replace `log.Fatal(err)` with `panic(err)`.
The other benefit of panic is that it's caught in our main function,
so other files can be checked even if this one contained something some garbage.

weeellz: Done.

lint/util.go
```diff
@@ -21,7 +20,7 @@ func collectFuncDecls(f *ast.File) []*ast.FuncDecl {
 func nodeString(fset *token.FileSet, x ast.Node) string {
 	var buf bytes.Buffer
 	if err := printer.Fprint(&buf, fset, x); err == nil {
```
fexolm: here should be err != nil

PR #31, opened 2018-05-09
lint/parenthesis-checker.go: add variable types check

fexolm:


lint/parenthesis-checker.go
```diff
@@ -27,21 +27,35 @@ func NewParenthesisChecker(ctx *Context) *ParenthesisChecker {
...
 		}
 	}
 	return c.warnings
 }
 
-func (c *ParenthesisChecker) validateResultDecl(f *ast.Field) {
+func (c *ParenthesisChecker) validateExpr(n ast.Node) {
 	// TODO improve suggestions for complex cases like (func([](func())))
+	// TODO improve linter output to write full type, not just place
+	// whereit could be simplified
```
Quasilyte: `s/whereit/where it/`

Quasilyte: We may want to disable `interfacer` in megacheck inside our CI script.
Method says `validateExpr`, but it accepts any `ast.Node`.
And that is not quite good, because `ast.Inspect` inside it will never handle anything except expressions.

This is not the first time `interfacer` bites us.
We should treat it as a suggestion, not obligation that breaks out build.

Quasilyte: What do you think about `s/validateExpr/validateType/`?
Types are expressions in AST terms, but not all expressions are types.

lint/parenthesis-checker.go
```diff
@@ -27,21 +27,35 @@ func NewParenthesisChecker(ctx *Context) *ParenthesisChecker {
...
+		case *ast.FuncDecl:
+			if decl.Type.Results == nil {
+				continue
+			}
+			for _, res := range decl.Type.Results.List {
+				c.validateExpr(res.Type)
+			}
+		case *ast.GenDecl:
+			for _, spec := range decl.Specs {
+				spec, ok := spec.(*ast.ValueSpec)
```
Quasilyte: This will panic for specs that are not `var` or `const`.
Add this to your tests and linter will (most likely) panic:
```go
type foo int
```

Quasilyte: What you may want to do for now is:
```go
if decl.Tok != token.VAR || decl.Tok != token.CONST {
	continue
}
```

testdata/parenthesis/checker_tests.go
```diff
@@ -7,3 +7,7 @@ func badReturn() [](func()) {
 func veryBadReturn() [](func([](func()))) {
 	return nil
 }
+
+var badVar [](func())
+
+var badVar2 [5](*int)
```
Quasilyte: Add at least one case for `const` and `type` declarations.

lint/parenthesis-checker.go
fexolm: SGTM, i ll rename it in a moment.
CI script is out of scope of that pr. I ll create issue for that.

lint/parenthesis-checker.go
Quasilyte: When you will handle `token.TYPE`, it's still required to skip `token.IMPORT`, as import specs are not interesting for your checker.

testdata/parenthesis/checker_tests.go
fexolm: @Quasilyte, could you please show, how this check could be used in type decl?

testdata/parenthesis/checker_tests.go
```diff
@@ -7,3 +7,15 @@ func badReturn() [](func()) {
...
+var _ [5](*int)
+var _ [](func())
+
+var (
+	_ int
+	_ [5](*int)
+	_ [](func())
+)
+
+const _ (int) = 5
```
Quasilyte: Add 1 type decl here and make sure that checker does not panic.
We will merge after that and you can continue to work on `ast.TypeSpec` in a separate PR.

testdata/parenthesis/checker_tests.go
Quasilyte:
```go
type foo [](func())
```
For `struct` types, you need to walk all fields.

lint/parenthesis-checker.go
```diff
@@ -27,21 +28,40 @@ func NewParenthesisChecker(ctx *Context) *ParenthesisChecker {
...
 }
 
-func (c *ParenthesisChecker) validateResultDecl(f *ast.Field) {
+func (c *ParenthesisChecker) validateType(n ast.Node) {
 	// TODO improve suggestions for complex cases like (func([](func())))
+	// TODO improve linter output to write full type, not just place
+	// where it could be simplified
 
-	ast.Inspect(f.Type, func(n ast.Node) bool {
+	fmt.Printf("%s\n", nodeString(c.ctx.FileSet, n))
```
Quasilyte: Missed this one.
`fmt.Printf("%s\n", nodeString(c.ctx.FileSet, n))` Should be removed.

Issue #37, opened 2018-05-09
all: create contribution guideline


fexolm:
We may need to add smth like contribution guideline which would formalize our workflaw.
For example: we may add label "in progress" if someone currently working on task and remove it when switch on the other task. 
It could save a lot of time if every team member would follow that rools.

Quasilyte:
Ok, makes sense.
We need `CONTRIBUTING.md` for sure.

Quasilyte:
We also need to formalize how we tag issues and most commit messages.
It's not 100% precise and I have doubts sometimes (like in situation with this issue title),
but if we end up with exact rules, it'll be easier.

fexolm:
@Quasilyte, closing?

Quasilyte:
You can close this and open more specific issues, if you like it more this way.
But in the current state, the document is not good enough.

Quasilyte:
@tengufromsky, if you have any issues figuring out how to contribute to this project, 
feel free to leave your questions here.

Quasilyte:
Related: #116.

PR #49, opened 2018-05-09
Start contributing guildline

fexolm:


CONTRIBUTING.org
```diff
@@ -0,0 +1,19 @@
+* Contributing to kfulint
+
+Here would be a set of guidelines that we follow when contributing to
+kfulint. You could propose changes in this document if you would like.
```
Quasilyte:
```
Here would be a set of guidelines that we follow when contributing to
kfulint. You could propose changes in this document if you would like.
```
Remove these lines and you can merge. :)

Quasilyte:
You can add "Refs #37" or "Updates #37" to commit message body.

PR #62, opened 2018-05-10
add test for parenthesis_checker

fexolm:


cmd/kfulint/testdata/parenthesis/checker_tests.go
```diff
@@ -5,6 +5,7 @@ func badReturn() [](func()) {
 	return nil
 }
 
+//TODO: cound simplify [](func[](func())) to []func([]func())
```
Quasilyte: s/cound/could.

PR #66, opened 2018-05-10
cmd/kfulint: add sanity checks

### LINKED ISSUE #57: cmd/kfulint: add sanity checks
Every linter must be properly handle files that may contain unexpected input.
For example:
```go
func external() // Function with nil body

func empty() {} // Function without input/output params and empty body

// Empty value/type specs:
var ()
const ()
type ()
```

And so on.

Quasilyte:
Add tests that may detect broken checkers.
Checkers that fail sanity checks are excluded
from testing list.

Fixes #57.


Quasilyte:
Hmmm, seems like we're only disabled top-level `testdata`.

PR #67, opened 2018-05-10
cmd/kfulint: use loader package to read, parse and typecheck

### LINKED ISSUE #54: cmd/kfulint: type checking error on stdlib packages
Many stdlib packages cause linter to give up with type checking error.
Most likely, it's because we run `Check` incorrectly (maybe we should consider build tags or something).

Here is output for `math` package:
```
./bin/kfulint -dir=$GOROOT/src/math
skip math: type check error: $GOROOTsrc/math/sincos_386.go:13:6: Sincos redeclared in this block
skip math_test: type check error: $GOROOTsrc/math/all_test.go:3557:7: undeclared name: SqrtGo
```
There are many other packages with similar problem:
- runtime (re-definitions of symbols)
- net (imports "C")
- net/http (undeclared name)

Other peculiar errors:
```
bufio:
scan_test.go:408:10: invalid operation: scanner (variable of type *bufio.Scanner) has no field or method MaxTokenSize

bytes:
bytes_test.go:577:31: undeclared name: EqualPortable

context:
context/x_test.go:29:58: undeclared name: XTestDeadlineExceededSupportsTimeout

crypto:
ase/cipher_ppc64le.go:27:6: encryptBlockAsm redeclared in this block
crypto/rand/rand_js.go:9:8: could not import syscall/js (can't find import: "syscall/js")
rc4/rc4_asm.go:13:18: method XORKeyStream already declared for type Cipher struct{s [256]uint32; i uint8; j uint8}
x509/root_darwin_arm_gen.go:38:6: main redeclared in this block
x509/root_cgo_darwin.go:221:8: could not import C (can't find import: "C")
```

These issues may be related or not.
This needs investigation.

Quasilyte:
This required to change dir-oriented approach to package-oriented.
kfulint now accepts package paths (well, import paths)
instead of directory paths.

Sanity tests are updated to import packages with fewer dependencies.
This helps to avoid big increase in test time.
(loader can be quite slow for packages with lots of imports.)

Fixes #54

### DIFF
--- a/cmd/kfulint/lint_test.go
+++ b/cmd/kfulint/lint_test.go
@@ -13,9 +13,14 @@ import (
 	"testing"
 )
 
-// binary is test linter executable name.
-// Using "exe" suffix to make it work on Windows as well.
-const binary = "testlint.exe"
+const (
+	// binary is test linter executable name.
+	// Using "exe" suffix to make it work on Windows as well.
+	binary = "testlint.exe"
+
+	// linterCmdPath holds full path to linter main pkg path.
+	linterCmdPath = "github.com/PieselBois/kfulint/cmd/kfulint/"
+)
 
 func TestMain(m *testing.M) {
 	// Before all tests are executed, we need to build linter first.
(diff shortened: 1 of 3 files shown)


Quasilyte:
This is not final version of course, just a way to make it possible to run checkers on any package that can be compiled (`loader` package does a lot of work for us).

PR #80, opened 2018-05-12
lint: add elseif checker v1

### LINKED ISSUE #77: lint: add if/else -> switch suggesting checker
If there is a chain of if/else, suggest switch as alternative.

These two functions should trigger the warning:
```go

func cond1(x int) string {
	if x == 0 {
		return "zero"
	} else if x < 0 {
		return "negative"
	} else {
		return "positive"
	}
}

func cond2(x int) string {
	if x == 0 {
		return "zero"
	} else if x < 0 {
		return "negative"
	}
	return "positive"
}
```
Suggested alternative is:
```go
func condSuggested(x int) string {
	switch {
	case x == 0:
		return "zero"
	case x < 0:
		return "negative"
	default:
		return "positive"
	}
}
```

TODO: find out if `if/else` is ever more readable or wanted instead of switch. 
Maybe init statements can complicate things.

Quasilyte:
Fixes #77

### DIFF
--- a/cmd/kfulint/lint_test.go
+++ b/cmd/kfulint/lint_test.go
@@ -43,6 +43,7 @@ var tests = []*struct {
 	{"parenthesis"},
 	{"param-duplication"},
 	{"underef"},
+	{"elseif"},
 }
 
 func runChecker(name, pkgPath string) (output []byte, err error) {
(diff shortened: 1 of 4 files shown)


Quasilyte:
@bontequero, could you review this for me?
We're better off merging recent checkers before new changes into architecture arrive.

PR #88, opened 2018-05-12
lint: make checkers unexported

weeellz:


cmd/kfulint/main.go
```diff
@@ -114,18 +114,9 @@ func (l *linter) InitContext() {
...
-		{"underef", lint.NewUnderefChecker(l.ctx)},
-		{"param-duplication", lint.NewParamDuplicationChecker(l.ctx)},
-	}
-
-	for _, c := range checkers {
-		// Nil enabledSet means "all checkers are enabled".
-		if l.enabledSet == nil || l.enabledSet[c.name] {
-			l.checkers = append(l.checkers, c)
+	for _, name := range lint.AvailableCheckers() {
+		if l.enabledSet == nil || l.enabledSet[name] {
```
Quasilyte: You've deleted meaningful comment here:
`// Nil enabledSet means "all checkers are enabled".`

lint/lint.go
```diff
@@ -38,3 +39,24 @@ type Context struct {
...
+
+var checkers = map[string]func(c *Context) Checker{
+	"param-name":        newParamNameChecker,
+	"type-guard":        newTypeGuardChecker,
+	"parenthesis":       newParenthesisChecker,
+	"underef":           newUnderefChecker,
+	"param-duplication": newParamDuplicationChecker,
+}
+
+func NewChecker(name string, ctx *Context) Checker {
```
Quasilyte: Build failed due to missing comment.
Add something like:
```
NewChecker returns checker that implements check of specified name.
```

lint/lint.go
```diff
@@ -38,3 +39,24 @@ type Context struct {
...
+	return checkers[name](ctx)
```
Quasilyte: This will panic if there is no checker `name`.
But that's OK for now.

lint/lint.go
```diff
@@ -38,3 +39,24 @@ type Context struct {
...
+}
+
+func AvailableCheckers() []string {
```
Quasilyte:
```
AvailableCheckers returns all check names that are supported.
Strings returned from this function can be safely used in NewChecker call.
```

Quasilyte:
LGTM when build is fixed (need to add doc comments).

cmd/kfulint/main.go
weeellz: Oops. Thanks.

lint/lint.go
weeellz: Done.

lint/lint.go
weeellz: Done.

lint/lint.go
weeellz: How to properly handle this?

Quasilyte: In normal circumstance, I would return `(Checker, error)`.
Mostly because it's a bad practice to do panic or fatal (exit) on constructor-like function error.

```go
if c := checkers[name]; c != nil {
	return c(ctx), nil
}
return nil, fmt.Errorf("checker %s not found", name)
```
The caller may decide what to do: if it's main, it could do `log.Fatalf("can't create checker: %v", err)`,
it it's testing function, it will just call `t.Errorf(...)`, and so on.

It's also OK to panic here because function contract requires it to be called only with `AvailableCheckers` returned values. Do what you think is better. There is no single right decision here, it depends on how we look at it.

Quasilyte: Handling such error is usually useful in tests, but I'm not sure we're interested in testing this particular function. It's either called properly, or it's a bad programming mistake and panic will lead to error location (stack trace).

weeellz: Thank you for detailed answer!

PR #89, opened 2018-05-12
add builtin functions shadowing checker

bontequero:


lint/builtin-shadow_checker.go
```diff
@@ -0,0 +1,66 @@
+package lint
+
+import (
+	"fmt"
+	"go/ast"
+)
+
+// BuiltinShadows detects when builtin functions shadowed in assignments
+type BuiltinShadows struct {
```
Quasilyte: Should be called `BuiltinShadowChecker`.

lint/builtin-shadow_checker.go
```diff
@@ -0,0 +1,66 @@
...
+		"imag":    struct{}{},
+		"len":     struct{}{},
+		"make":    struct{}{},
+		"new":     struct{}{},
+		"panic":   struct{}{},
+		"print":   struct{}{},
+		"println": struct{}{},
+		"real":    struct{}{},
+		"recover": struct{}{},
+	}
```
Quasilyte: If you don't like repeating `struct{}{}`, there are 2 options:
1. Use `map[string]bool`. That's less ugly.
2. Use this idiom:
```go
	builtins := []string{
		"append",
		"cap",
		"close",
		"complex",
		"copy",
		"delete",
		"imag",
		"len",
		"make",
		"new",
		"panic",
		"print",
		"println",
		"real",
		"recover",
	}
	builtinFuncs := make(map[string]struct{})
	for i := range builtins {
		builtinFuncs[builtins[i]] = struct{}{}
	}
```
But I'm OK with current approach.

Quasilyte: It's also better to initialize map literal inside constructor and store it inside checker.
This way, each `Check` call doesn't have to construct that map again.

Quasilyte: See `param-name` checker for example. It stores small map inside it.

lint/builtin-shadow_checker.go
```diff
@@ -0,0 +1,53 @@
+package lint
+
+import (
+	"go/ast"
+)
+
+// builtinShadowCheck detects when builtin functions shadowed in assignments
```
Quasilyte: This sentence misses period.
Could you add missing punctuation?

PR #91, opened 2018-05-13
lint: get rid of []Warning inside every checker

### LINKED ISSUE #81: lint: get rid of []Warning inside every checker
Add `Warning` channel or mutex-guarded slice to `Context` and use it from each checker instead of storing `[]Warning` in every checker.

fexolm:
fixes #81


cmd/kfulint/main.go
```diff
@@ -107,13 +108,13 @@ func (l *linter) LoadProgram() {
 	l.prog = prog
 }
 
-func (l *linter) InitContext() {
+func (l *linter) initContext() {
```
Quasilyte: Usually, capitalized (exportes) methods on unexported types are OK to be without a comment, and go lint does not trigger on them. The reason it's this way is to emphasize that those methods are expected to be called from outside of the type, as opposed to unexported methods that should be called only inside type itself (like private methods).
I think we can even write a checker for this: if unexported type method is ever called outside of it's own methods, suggest making it "exported", otherwise suggest making it "unexported" (if it's only called from within that type).
This check can be disabled by people who don't like this convention, but it may help for projects that try to comply with it.

The main benefit is: when you're making that type exported, your API is already almost done, what is left is to add comments to exported methods. If all methods are unexported, you will need to do a manual job.

lint/lint.go
```diff
@@ -28,6 +29,17 @@ type Context struct {
 
 	// TypesInfo carries parsed packages types information.
 	TypesInfo *types.Info
+
+	// Warnings contains warnings from all checkers
+	Warnings []Warning
+
+	mutex sync.Mutex
```
Quasilyte: There is a convention to place mutex that protects struct fields above those fields with a special comment.
Do something like this:
```go
mutex sync.Mutex // Protects the rest of the fields

// Warnings contains warnings from all checkers
Warnings []Warning
```

You can find many examples in stdlib. Do a grep from `GOROOT/src` like this:
```
grep -nr -C 3 'sync.Mutex'
```

Another trick is to use anonymous struct that embeds mutex and contains named "guarded" fields.
It's overkill for this case as we may change the exact sync patterns and maybe move to checker-local storage again (but this time, no warnings inside checker itself) to avoid any sync at all.

Quasilyte:
This makes transition to something better easier and achieves a goal of moving []Warning outside of checkers. You can merge it when you think it's ready.

Quasilyte:
Also checkout new @weeellz patch, it may require changes before this can be merged.

lint/lint.go
```diff
@@ -60,7 +60,7 @@ var checkers = map[string]func(c *Context) Checker{
 	"param-duplication": newParamDuplicationChecker,
 }
 
-// NewChecker returns checker that implements check of specified name.
+// NewChecker rnnneturns checker that implements check of specified name.
```
Quasilyte: This seems not right...
`s/rnnneturns/returns/`

Quasilyte: 1ac4953f7a3a63c8bc6a095b8bae4144b229f9b2

Issue #46, opened 2018-05-09
lint/underef_checker.go: add array support


fexolm:
We need to add functionality to unwrap expressions like:
`(*a)[5]` ===> `a[5]` 
Note: it should be done just for arrays, not slices.

Quasilyte:
This will definitely require working with `types.Info`. :)

fexolm:
Who knows :D

PR #94, opened 2018-05-14
lint: underef gives false positives for multiple levels of indirection

### LINKED ISSUE #68: lint: underef gives false positives for multiple levels of indirection
For types like `**T` and `***T` underef suggest to remove dereferencing even though it will lead to invalid code. It should take types into account.

Examples:
```go
	type point struct{ x, y int }
	pt := point{}
	pt1 := &pt
	pt2 := &pt1
	pt3 := &pt2
	pt4 := &pt3
	
	_ = (*pt2).x
	_ = (**pt3).x
	_ = (***pt4).x
```

There can be more cases where it gives false positives, but this one can be resolved in isolation.

fexolm:
fix #68

### DIFF
--- a/cmd/kfulint/testdata/underef/checker_tests.go
+++ b/cmd/kfulint/testdata/underef/checker_tests.go
@@ -46,16 +46,10 @@ func multipleIndir1() {
 	pt3 := &pt2
 	pt4 := &pt3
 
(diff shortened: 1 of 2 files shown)


fexolm:
also fixes #46

Issue #47, opened 2018-05-09
lint/parenthesis_checker.go: add support for types 


fexolm:
we neet to add support for types, declareted like so:
`type a (int)` ===>> `type a int`

Quasilyte:
Hey, that's not type aliases. :)
Go have "real" type aliases, and they look like this:
```go
type a = b
```
https://golang.org/ref/spec#Alias_declarations

fexolm:
Oh, my mistake.

PR #95, opened 2018-05-14
lint: parenthesis suggests to simplify size expressions inside array types

### LINKED ISSUE #60: lint: parenthesis suggests to simplify size expressions inside array types
```
cmd/internal/obj/x86/asm6.go:2043:14: parenthesis: could simplify (ALAST + 1) to ALAST + 1
```

`[(A + B)]int` is an array of ints with size `A+B`. 
Expressions can be complex and it's not quite right to trigger on them.

I don't remember any other cases except for array types, so maybe all what is required is to avoid array type length checks.

```go
	ArrayType struct {
		Lbrack token.Pos // position of "["
		Len    Expr      // Ellipsis node for [...]T array types, nil for slice types
		Elt    Expr      // element type
	}
```

`parenthesis` should not visit `ArrayType.Len`.

fexolm:
fixes #60

### DIFF
--- a/cmd/kfulint/testdata/parenthesis/checker_tests.go
+++ b/cmd/kfulint/testdata/parenthesis/checker_tests.go
@@ -31,7 +31,7 @@ var (
 ///: could simplify (int) to int
 const _ (int) = 5
 
-//TODO: could simplify (int) to int
+///: could simplify (int) to int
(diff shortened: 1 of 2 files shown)


fexolm:
also fixes #47

PR #96, opened 2018-05-14
lint: long chain expression checker v1

### LINKED ISSUE #11: lint: add long chain expression checker
Find long expression chains that are repeated more than N times (some kind of threshold) and consist of multiple selector/index expressions.

So, this code will trigger a warning:
```go
	type something struct {
		with struct {
			very struct {
				big struct {
					nesting struct {
						value int
					}
				}
			}
		}
	}
	var x something

	switch {
	case x.with.very.big.nesting.value > 0:
		fmt.Printf("positive value: %d\n", x.with.very.big.nesting.value)
	case x.with.very.big.nesting.value < 0:
		fmt.Printf("negative value: %d\n", x.with.very.big.nesting.value)
	default:
		fmt.Println("zero")
	}
```
And the alternative we are proposing:
```go
	value := x.with.very.big.nesting
	switch {
	case value > 0:
		fmt.Printf("positive value: %d\n", value)
	case value < 0:
		fmt.Printf("negative value: %d\n", value)
	default:
		fmt.Println("zero")
	}
```
Or, even better:
```go
	switch value := x.with.very.big.nesting; {
	case value > 0:
		fmt.Printf("positive value: %d\n", value)
	case value < 0:
		fmt.Printf("negative value: %d\n", value)
	default:
		fmt.Println("zero")
	}
```

We don't have to provide "fixed" code snippet.
The minimal goal is to alert the possibility for code improvement.
It means that in the simplest case it's OK to just report the problem, like:
> Expression chain `x.with.very.big.nesting.value` repeated 4 times, consider assigning it to local variable

fexolm:
fixes #11

### DIFF
--- a/cmd/kfulint/testdata/long-chain/checker_tests.go
+++ b/cmd/kfulint/testdata/long-chain/checker_tests.go
@@ -10,6 +10,22 @@ type t struct {
 					}
 				}
 			}
+
+			c1 struct {
+				d struct {
+					e struct {
+						p1, p2, p3, p4 int
+					}
+				}
+			}
+
+			c2 struct {
+				d struct {
+					e struct {
+						p1, p2, p3, p4 int
+					}
+				}
+			}
 		}
 	}
 }
@@ -23,5 +39,11 @@ func sample() {
 	case b.a.b.c.d.e.p2:
 	case b.a.b.c.d.e.p3:
 	case b.a.b.c.d.e.p4:
+	}
+
+	switch a {
+	case b.a.b.c.d.e.p1:
+	case b.a.b.c1.d.e.p1:
+	case b.a.b.c2.d.e.p1:
 	}
 }
--- a/cmd/kfulint/testdata/long-chain/checker_tests.go
+++ b/cmd/kfulint/testdata/long-chain/checker_tests.go
@@ -1,0 +1,27 @@
+package checker_test
+
+type t struct {
+	a struct {
+		b struct {
+			c struct {
+				d struct {
+					e struct {
+						p1, p2, p3, p4 int
+					}
+				}
+			}
+		}
+	}
+}
+
+func sample() {
+	a := 1
+	b := t{}
+	///: Expression chain b.a.b.c.d.e repeated multiple times consider assigning it to local variable
+	switch a {
+	case b.a.b.c.d.e.p1:
+	case b.a.b.c.d.e.p2:
+	case b.a.b.c.d.e.p3:
+	case b.a.b.c.d.e.p4:
+	}
+}
--- a/cmd/kfulint/testdata/long-chain/checker_tests.go
+++ b/cmd/kfulint/testdata/long-chain/checker_tests.go
@@ -2,28 +2,26 @@ package checker_test
 
 type t struct {
 	a struct {
-		b struct {
-			c struct {
-				d struct {
-					e struct {
-						p1, p2, p3, p4 int
-					}
+		c struct {
+			d struct {
+				e struct {
+					p1, p2, p3, p4 int
 				}
 			}
+		}
 
-			c1 struct {
-				d struct {
-					e struct {
-						p1, p2, p3, p4 int
-					}
+		c1 struct {
+			d struct {
+				e struct {
+					p1, p2, p3, p4 int
 				}
 			}
+		}
 
-			c2 struct {
-				d struct {
-					e struct {
-						p1, p2, p3, p4 int
-					}
+		c2 struct {
+			d struct {
+				e struct {
+					p1, p2, p3, p4 int
 				}
 			}
 		}
@@ -33,17 +31,17 @@ type t struct {
 func sample() {
 	a := 1
 	b := t{}
-	///: Expression chain b.a.b.c.d.e repeated multiple times consider assigning it to local variable
+	///: Expression chain b.a.c.d.e repeated multiple times consider assigning it to local variable
 	switch a {
-	case b.a.b.c.d.e.p1:
-	case b.a.b.c.d.e.p2:
-	case b.a.b.c.d.e.p3:
-	case b.a.b.c.d.e.p4:
+	case b.a.c.d.e.p1:
+	case b.a.c.d.e.p2:
+	case b.a.c.d.e.p3:
+	case b.a.c.d.e.p4:
 	}
 
 	switch a {
-	case b.a.b.c.d.e.p1:
-	case b.a.b.c1.d.e.p1:
-	case b.a.b.c2.d.e.p1:
+	case b.a.c.d.e.p1:
+	case b.a.c1.d.e.p1:
+	case b.a.c2.d.e.p1:
 	}
 }
--- a/cmd/kfulint/testdata/parenthesis/checker_tests.go
+++ b/cmd/kfulint/testdata/parenthesis/checker_tests.go
@@ -48,3 +48,7 @@ type myString string
 func convertPtr(x string) *myString {
 	return (*myString)(&x)
 }
+
+func multipleReturn() (int, bool) {
+	return 1, true
+}
--- a/lint/long-chain_checker.go
+++ b/lint/long-chain_checker.go
@@ -68,29 +68,28 @@ func (c *longChainChecker) exprToList(expr ast.Expr) []ast.Expr {
 			tmp = t.X
 		default:
 			res = append(res, tmp)
-			res = reverseExprSlice(res)
+			reverseExprSlice(res)
 			return res
 
 		}
 	}
 }
 
-func reverseExprSlice(e []ast.Expr) []ast.Expr {
+func reverseExprSlice(e []ast.Expr) {
 	for i := len(e)/2 - 1; i >= 0; i-- {
 		j := len(e) - 1 - i
 		e[i], e[j] = e[j], e[i]
 	}
-	return e
 }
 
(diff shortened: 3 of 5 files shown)


cmd/kfulint/testdata/long-chain/checker_tests.go
```diff
@@ -0,0 +1,27 @@
...
+					e struct {
+						p1, p2, p3, p4 int
+					}
+				}
+			}
+		}
+	}
+}
+
+func sample() {
```
Quasilyte: Can you add samples that do not trigger a warning?
For example, chains with `n=2` and `n=3`.

cmd/kfulint/testdata/long-chain/checker_tests.go
```diff
@@ -24,4 +40,10 @@ func sample() {
 	case b.a.b.c.d.e.p3:
 	case b.a.b.c.d.e.p4:
 	}
+
+	switch a {
+	case b.a.b.c.d.e.p1:
+	case b.a.b.c1.d.e.p1:
+	case b.a.b.c2.d.e.p1:
```
Quasilyte: I guess simpler sequences like `b.a.b` are not reported too, so it's OK.
Thank you.

lint/long-chain_checker.go
```diff
@@ -0,0 +1,119 @@
...
+		case *ast.SelectorExpr:
+			res = append(res, t.Sel)
+			tmp = t.X
+		default:
+			res = append(res, tmp)
+
+			for i := len(res)/2 - 1; i >= 0; i-- {
+				opp := len(res) - 1 - i
+				res[i], res[opp] = res[opp], res[i]
+			}
```
Quasilyte: I think it's copy/pasted slice reverse code.
I don't like `opp` name at all. Can it be `j`?
In addition, I would like it to be moved to `reverseExprSlice` method just to make intention more clear.

cmd/kfulint/testdata/long-chain/checker_tests.go
Quasilyte: We should either add a todo or create an issue to add more repeated long chains like `a.b[i][j].c` to make sure that all things we like to detect are reported.
(For now, it's good enough to match only selector expressions.)

lint/long-chain_checker.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+}
+
+func (c *longChainChecker) checkSwitch(stmt *ast.SwitchStmt) {
+	exprs := []ast.Expr{}
+	for _, s := range stmt.Body.List {
+		cas := s.(*ast.CaseClause)
+		if cas.List == nil {
+			continue
+		}
```
Quasilyte: It's actually OK to append nil or empty slice.
Maybe we can remove this if statement completely?
https://play.golang.org/p/tca2ImY64RJ

Quasilyte:
I'll take a closer at evening. 
Thanks for doing this, by the way.

Can't wait to check it on a bigger code base.
Have you tried to run it with `lintwalk` on `$GOOROOT/src`?
(By the way, it can help to reveal bugs.)

lint/long-chain_checker.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	const n = 3
+
+	if len(cp) > n {
+		c.warn(cp, stmt)
+	}
+}
+
+func (c *longChainChecker) exprToList(expr ast.Expr) []ast.Expr {
```
Quasilyte: This method is rather hard to understand.
Can you either add comments that describe what (and, more importantly, why) you are doing here
or simplify it? The hardest part is proving it's correct.
It seems like it collects selector expression `X`s, but what if there are non-selector expressions along the way?
Like this: `x.y[i].z()`.

lint/long-chain_checker.go
```diff
@@ -0,0 +1,121 @@
...
+		return a
+	}
+	return b
+}
+
+func (c *longChainChecker) warn(exprs []ast.Expr, stmt ast.Stmt) {
+	var s []string
+	for _, e := range exprs {
+		s = append(s, nodeString(c.ctx.FileSet, e))
+	}
```
Quasilyte: Not very performance-critical application and method, but here is advice that could be useful in future:
when you know exact length of slice you want to fill, it's possible to reserve the memory once.
```diff
+	var s []string
+	for _, e := range exprs {
+		s = append(s, nodeString(c.ctx.FileSet, e))
+	}
-	s := make([]string, len(exprs))
-	for i, e := range exprs {
-		s[i] = nodeString(c.ctx.FileSet, e)
-	}
```
Also note that you can avoid append calls with direct index assignment.
If you want to keep append there, `len(exprs)` should be used as a `cap` argument, not `len` argument:
```go

s := make([]string, 0, len(exprs))
for _, e := range exprs {
	s = append(s, nodeString(c.ctx.FileSet, e))
}
```
Note the `0` length.

Quasilyte:
If you have time, try running this checker with lower `n` and see if it finds issues in `$GOROOT/src`.
This could help us to find false positives and/or crashes.

lint/long-chain_checker.go
```diff
@@ -0,0 +1,121 @@
...
+		j := len(e) - 1 - i
+		e[i], e[j] = e[j], e[i]
+	}
+	return e
+}
+
+func (c *longChainChecker) commonPrefix(a, b []ast.Expr) []ast.Expr {
+	res := []ast.Expr{}
+
+	len := c.min(len(a), len(b))
```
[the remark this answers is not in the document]
Quasilyte: The `len` is probably the most frequent case for builtin function shadowing.
Probably we should suggest alternatives like: `l` or `length`.

lint/long-chain_checker.go
```diff
@@ -0,0 +1,121 @@
...
+
+func reverseExprSlice(e []ast.Expr) []ast.Expr {
+	for i := len(e)/2 - 1; i >= 0; i-- {
+		j := len(e) - 1 - i
+		e[i], e[j] = e[j], e[i]
+	}
+	return e
+}
+
+func (c *longChainChecker) commonPrefix(a, b []ast.Expr) []ast.Expr {
```
Quasilyte: It's better to call these `x` and `y`.
Even better, `xs` and `ys`.

lint/long-chain_checker.go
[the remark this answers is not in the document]
Quasilyte: By the way, in this particular case, you can avoid explicit length and use range loop if you replace `min` with something that helps you to make `xs` shorter and `ys` longer, then you do `for i, x := range xs {` and use `i` to assess `y` with `ys[i]`.

lint/long-chain_checker.go
fexolm: the idea is to find the most long prefix, which would fit all expressions.

lint/long-chain_checker.go
fexolm: Good advice, thx

lint/long-chain_checker.go
fexolm: ok

lint/long-chain_checker.go
```diff
@@ -0,0 +1,121 @@
...
+	res := []ast.Expr{}
+	tmp := expr
+	for {
+		switch t := tmp.(type) {
+		case *ast.SelectorExpr:
+			res = append(res, t.Sel)
+			tmp = t.X
+		default:
+			res = append(res, tmp)
+			res = reverseExprSlice(res)
```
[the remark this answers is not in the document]
Quasilyte: And by the way, it's better to think of slices as a struct like:
```
type slice struct {
  data *T
  length int
  capacity int
}
```
There are no "references" in Go terminology.
If you look into slice struct definition above, it becomes obvious why you don't need to return input slice in case length/capacity didn't changed.
The data (pointer to be first element) is modified - that's what matters.

lint/long-chain_checker.go
fexolm: fixed

lint/long-chain_checker.go
Quasilyte: Comment should go to the source file, not here. :)
We are (hopefully) not the only maintainers of this linter, so it should be as accessible for others as possible.

cmd/kfulint/testdata/parenthesis/checker_tests.go
```diff
@@ -48,3 +48,7 @@ type myString string
 func convertPtr(x string) *myString {
 	return (*myString)(&x)
 }
+
+func multipleReturn() (int, bool) {
```
Quasilyte: This looks unrelated, but OK. :)

lint/long-chain_checker.go
```diff
@@ -0,0 +1,139 @@
...
+// The general idea is to find greatest common prefix of case expression
+// statements. To do that, we split each case expression by "." (only for
+// SelectorExpr) and compare each ith element of every case expression.
+// If they all equal, we continue algorithm on i+1. If they are not
+// we assume prefix length = i-1
+// If prefix length greater then n we create warning
+//
+// TODO: warn if at least m case expressions have common prefix with
+// length > n
+
```
Quasilyte: This empty line makes comment detached, so it's not a proper doc comment from AST/godoc point of view.
It's OK for this revision. We can fix that later. This review is already gone for too long.

Quasilyte: It's better to have misplaced comment than no documentation at all. :)

PR #98, opened 2018-05-17
cmd/kfulint: add warnings if enabled package not found

fexolm:
fixes #98

### DIFF
--- a/cmd/kfulint/main.go
+++ b/cmd/kfulint/main.go
@@ -8,7 +8,6 @@ import (
 	"go/types"
 	"log"
 	"os"
-	"regexp"
 	"runtime"
 	"strings"
 	"sync"
@@ -31,8 +30,8 @@ type linter struct {
 
 	// Command line flags:
 
-	packages   []string
-	enabledSet map[string]bool
+	packages        []string
+	enabledCheckers []string
 }
 
 func main() {
@@ -77,20 +76,11 @@ func parseArgv(l *linter) {
 	case "":
 		// Empty set. Semantically "disable-all".
 		// Can be used to run all pipelines without actual checkers.
-		l.enabledSet = map[string]bool{}
+		l.enabledCheckers = []string{}
 	default:
 		// Comma-separated list of names.
-		l.enabledSet = make(map[string]bool)
-		nameRE := regexp.MustCompile(`[a-z][a-z0-9_\-]*`)
-		for i, s := range strings.Split(*enable, ",") {
-			s = strings.TrimSpace(s)
-			l.enabledSet[s] = true
-			if !nameRE.MatchString(s) {
-				log.Fatalf("-enable element #%d: invalid %q checker name", i, s)
-			}
-		}
+		l.enabledCheckers = strings.Split(*enable, ",")
 	}
-
 }
 
 func (l *linter) LoadProgram() {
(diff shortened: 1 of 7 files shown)


cmd/kfulint/main.go
```diff
@@ -123,11 +113,20 @@ func (l *linter) LoadProgram() {
...
 func (l *linter) InitCheckers() {
-	for _, name := range lint.AvailableCheckers() {
-		// Nil enabledSet means "all checkers are enabled".
-		if l.enabledSet == nil || l.enabledSet[name] {
-			l.checkers = append(l.checkers, checker{name, lint.NewChecker(name, l.ctx)})
+	if l.enabledCheckers == nil {
+		for _, name := range lint.AvailableCheckers() {
+			c, _ := lint.NewChecker(name, l.ctx)
+			l.checkers = append(l.checkers, checker{name, c})
+		}
```
Quasilyte: Maybe add `return` after this if body?
It's understandable that for loop above will not execute any code in case `l.enabledCherkers == nil`, but
if there ever be other code, things will get ugly.
We should be explicit here.

cmd/kfulint/main.go
```diff
@@ -123,11 +113,20 @@ func (l *linter) LoadProgram() {
...
+	}
+
+	for _, s := range l.enabledCheckers {
+		name := strings.TrimSpace(s)
+		c, ok := lint.NewChecker(strings.TrimSpace(name), l.ctx)
```
Quasilyte: I think double trimming is a mistake.

cmd/kfulint/main.go
fexolm: Ok

cmd/kfulint/main.go
fexolm: It is first time where it trimmed

lint/lint.go
```diff
@@ -67,8 +67,12 @@ var checkers = map[string]func(c *Context) Checker{
 }
 
 // NewChecker returns checker that implements check of specified name.
-func NewChecker(name string, ctx *Context) Checker {
-	return checkers[name](ctx)
+func NewChecker(name string, ctx *Context) (Checker, bool) {
```
Quasilyte: I think you have considered returning `nil` here, right?
Could you elaborate why `nil, false` is better than just `nil`? (we still return "dangerous" nil)

cmd/kfulint/main.go
Quasilyte: `name := strings.TrimSpace(s)`
Then `lint.NewChecker(strings.TrimSpace(name), l.ctx)`.

Quasilyte: `name` is trimmed twice.

fexolm: oh, i see. sry

lint/lint.go
fexolm: It just looks easier to understand for me:)
But yes, we could use nil here if u want so

Quasilyte: You should return `error` instead of `bool` here, like:
```go
return nil, errors.New("checker not found")
```
Otherwise we're leaking that checkers are a map and (nil, false) basically means entry is missing. If there are ever 2 kinds of errors, this `bool` couldn't express that.

On the caller side:
```go
c, err := lint.NewChecker(name, l.ctx)
if err != nil {
	log.Fatalf("%s: %v", err)
}
```

Quasilyte: I'm okay with 2-value return, but `error` is more usual way to represent failure.
`,ok` is also fine for now, though.
You can decide here on your own.

fexolm: Oh, I see. Thanks.

fexolm: Done

PR #101, opened 2018-05-19
lint: add switch -> if suggestions for 1-case statements

weeellz:
Works for all kind of switch.

### DIFF
--- a/lint/switchif_checker.go
+++ b/lint/switchif_checker.go
@@ -1,0 +1,52 @@
+package lint
+
+import (
+	"fmt"
+	"go/ast"
+)
+
+type switchifChecker struct {
+	ctx *Context
+}
+
+func newSwitchifChecker(ctx *Context) Checker {
+	return &switchifChecker{ctx: ctx}
+}
+
+func (c *switchifChecker) Check(f *ast.File) {
+	ast.Inspect(f, func(x ast.Node) bool {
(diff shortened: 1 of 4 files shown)


weeellz:
@Quasilyte, should checker suggest alternative?

Quasilyte:
@weeellz, I think we will have a good documentation one day that will explain how to address every warning and what alternatives are implied. Maybe we can suggest alternatives with special kfulint flag or with other command.
In any case, it's out of scope of our current tasks.

I'll do a review tomorrow if you don't mind.

weeellz:
@Quasilyte, ok.

lint/switchif_checker.go
```diff
@@ -0,0 +1,52 @@
...
+	ctx *Context
+}
+
+func newSwitchifChecker(ctx *Context) Checker {
+	return &switchifChecker{ctx: ctx}
+}
+
+func (c *switchifChecker) Check(f *ast.File) {
+	ast.Inspect(f, func(x ast.Node) bool {
+		if stmt, ok := x.(*ast.SwitchStmt); ok {
```
fexolm: You could use typeswitch here

fexolm:
``` 
switch stmt.(type) {
   case *ast.SwitchStmt:
       c.checkSwitchStmt(stmt, stmt.Body)
   ...
}

PR #102, opened 2018-05-20
lint: add checker for unexported method calls on unexported types

### LINKED ISSUE #92: lint: add checker for unexported method calls on unexported types
Calling unexported method on unexported type is not an error, but it's nicer
to make methods that are intended to be called from outside of type itself exported (`golint` doesn't require having doc-comments for them, so it doesn't add a burden).

This makes refactoring a type into exported later easier.

This code triggers a warning:
```go
package main

type foo struct{}

func (foo) unexported() {}
func (foo) Exported() {}

func main() {
  var f foo
  f.unexported() // Warning
  f.Exported() // OK, call of exported method
}
```

fexolm:
fixes #92

### DIFF
--- a/cmd/kfulint/testdata/unexported-call/checker_tests.go
+++ b/cmd/kfulint/testdata/unexported-call/checker_tests.go
@@ -9,9 +9,9 @@ func getFoo() foo { return foo{} }
 
 func f() {
 	var fo foo
-	///: getFoo().unexported() should be exported
+	/// getFoo().unexported() should be exported
 	getFoo().unexported()
-	///: fo.unexported() should be exported
+	/// fo.unexported() should be exported
 	fo.unexported()
 	fo.Exported()
 }
--- a/cmd/kfulint/testdata/unexported-call/checker_tests.go
+++ b/cmd/kfulint/testdata/unexported-call/checker_tests.go
@@ -1,0 +1,17 @@
+package checker_test
+
+type foo struct{}
+
+func (f foo) unexported() int { return 1 }
+func (f foo) Exported() int   { return 1 }
+
+func getFoo() foo { return foo{} }
+
+func f() {
+	var fo foo
+	///: getFoo().unexported() should be exported
+	getFoo().unexported()
+	///: fo.unexported() should be exported
+	fo.unexported()
+	fo.Exported()
+}
--- a/lint/unexported-call_checker.go
+++ b/lint/unexported-call_checker.go
@@ -9,7 +9,7 @@ import (
 type unexportedCallChecker struct {
 	baseLocalExprChecker
 
-	funcName string
+	recvName string
 }
 
 func unexportedCallCheck(ctx *context) func(*ast.File) {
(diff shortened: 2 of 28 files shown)


cmd/kfulint/testdata/unexported-call/checker_tests.go
```diff
@@ -0,0 +1,17 @@
...
+func getFoo() foo { return foo{} }
+
+func f() {
+	var fo foo
+	///: getFoo().unexported() should be exported
+	getFoo().unexported()
+	///: fo.unexported() should be exported
+	fo.unexported()
+	fo.Exported()
+}
```
Quasilyte: Need a test case for `foo` method that calls unexported methods and gets no warnings.

lint/unexported-call_checker.go
```diff
@@ -0,0 +1,79 @@
...
+	for _, decl := range collectFuncDecls(f) {
+		if decl.Body == nil {
+			continue
+		}
+		name := ""
+		cond := decl.Recv != nil &&
+			decl.Recv.List != nil &&
+			len(decl.Recv.List) == 1 &&
+			decl.Recv.List[0].Names != nil &&
+			len(decl.Recv.List[0].Names) == 1
```
Quasilyte: I believe two nil checks for `List` fields are redundant as nil slice length is 0.
The `== 1` check would cover both cases and is perfectly safe.

Quasilyte: The `gosimple` doesn't do a best job with it's [S1009](https://staticcheck.io/docs/gosimple#S1009).
It should catch these cases as well.

lint/unexported-call_checker.go
```diff
@@ -0,0 +1,79 @@
...
+	typ := c.ctx.TypesInfo.TypeOf(sel.Sel)
+	sig, ok := typ.(*types.Signature)
+	if !ok {
+		return
+	}
+
+	if sig.Recv() == nil || sig.Recv().Type() == nil {
+		return
+	}
+	typ2, ok := sig.Recv().Type().(*types.Named)
```
Quasilyte: Maybe `s/typ2/recvTyp/`?

Quasilyte:
You've said that it handles GOROOT without crashes?
If that's true, LGTM when comments are addressed.

fexolm:
@will merge it after #109

lint/unexported-call_checker.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"fmt"
+	"go/ast"
+	"go/types"
+)
+
+type unexportedCallChecker struct {
+	baseLocalExprChecker
+
+	funcName string
```
Quasilyte: Should be `recvName`.

lint/unexported-call_checker.go
```diff
@@ -0,0 +1,74 @@
...
+	if cond {
+		c.funcName = decl.Recv.List[0].Names[0].Name
+	}
+	return true
+}
+
+// Check finds calls of unexported method from unexported type
+// outside that type.
+//
+//TODO: update description and warning message
```
Quasilyte: Please put a spece after `//`.
Comments without a space are treated as "pragmas" in Go gc toolchain.
Especially if "keyword" ends with `:`.

fexolm:
@Quasilyte, fixed.

PR #104, opened 2018-05-21
lint/param-duplication_checker.go improve suggestions 

### LINKED ISSUE #48: lint/param-duplication_checker.go improve suggestions
Right now for 
`func(a int, b int, c int) `
we would output 2 suggestions
`a, b int`
`b, c int`
but it should be just one
`a, b, c int`

fexolm:
fixes #48

### DIFF
--- a/lint/param-duplication_checker.go
+++ b/lint/param-duplication_checker.go
@@ -41,31 +41,46 @@ func (c *ParamDuplicationChecker) checkParamDuplication(decl *ast.FuncDecl) {
 	if len(params) < 2 {
 		return
 	}
+	equalRange := [][]*ast.Field{}
+	equalRange = append(equalRange, []*ast.Field{params[0]})
+	warn := false
 	for i, p := range params[1:] {
 		if astcmp.EqualExpr(p.Type, params[i].Type) {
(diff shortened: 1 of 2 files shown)


Quasilyte:
Nice. I'll take a look this evening.

lint/param-duplication_checker.go
```diff
@@ -41,31 +41,46 @@ func (c *ParamDuplicationChecker) checkParamDuplication(decl *ast.FuncDecl) {
 	if len(params) < 2 {
 		return
 	}
+	equalRange := [][]*ast.Field{}
+	equalRange = append(equalRange, []*ast.Field{params[0]})
```
Quasilyte: What do you think about:
```diff
-	equalRange := [][]*ast.Field{}
-	equalRange = append(equalRange, []*ast.Field{params[0]})
+	equalRange := [][]*ast.Field{{params[0]}}
```

Quasilyte:
The sequence of equal ranges idea looks good.

The combined error is even better than I've anticipated. It seems user can simple copy and paste suggestion to fix the whole signature. That's neat.

PR #114, opened 2018-05-26
lint: add unslice_checker

ludweeg:


lint/unslice_checker.go
```diff
@@ -0,0 +1,38 @@
+package lint
+
+import (
+	"go/ast"
+	"go/types"
+)
+
+// unsliceChecker finds slice expressions that can be
+// simplified to sliced expression.
+func unsliceCheck(ctx *context) func(*ast.File) {
```
Quasilyte: We forgot to update function comment after function rename.
Comment says `unsliceChecker` when it should be `unsliceCheck`.

Quasilyte:
You can merge it in this state or update a comment.

PR #115, opened 2018-05-26
lint: add checks for uninformative comments

weeellz:


lint/comments_checker.go
```diff
@@ -0,0 +1,33 @@
...
+				if c.regexp.MatchString(decl.Doc.List[0].Text) {
+					c.warn(decl)
+				}
+			}
+		}
+	}
+}
+
+func (c *commentsChecker) warn(decl *ast.FuncDecl) {
+	c.ctx.Warn(decl, "comment should contain function name")
```
Quasilyte: What about saying `silencing go lint doc-comment warnings is unadvised`?
Or something that explicitly tells that we detected a comment that tries to avoid go lint notice, but doing so is a bad practice.

lint/comments_checker.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func (c *commentsChecker) CheckFile(f *ast.File) {
+	for _, decl := range f.Decls {
+		if decl, ok := decl.(*ast.FuncDecl); ok {
+			if decl.Doc != nil {
+				if c.regexp.MatchString(decl.Doc.List[0].Text) {
+					c.warn(decl)
+				}
+			}
```
Quasilyte: You can do something like this:
```go
- if decl.Doc != nil {
- 	if c.regexp.MatchString(decl.Doc.List[0].Text) {
- 		c.warn(decl)
- 	}
- }
+ if decl.Doc != nil && c.regexp.MatchString(decl.Doc.List[0].Text) {
+ 	c.warn(decl)
+ }
```
But this is a matter of taste and preference.

We can ask what @fexolm prefers here. :)

lint/comments_checker.go
weeellz: Ok.

lint/comments_checker.go
fexolm: I prefer second one if line wouldn't be so long.
If so, you could use smth like
```
cond := decl.Doc != nil &&
        c.regexp.MatchString(...)
if cond {
...
}
```

lint/comments_checker.go
```diff
@@ -0,0 +1,33 @@
+package lint
+
+import (
+	"go/ast"
+	"regexp"
+)
+
+type commentsChecker struct {
+	ctx    *context
+	regexp *regexp.Regexp
```
fexolm: maybe s/regexp/re It conflicts with "regexp" package name.

Quasilyte: It can also describe actual regexp and what it supposed to match.
For example: `badCommentRE` or `commentRE`.
But that's up to you. Naming things is hard.

By the way, conflicts at field names are not a problem as you access them through object and there is no ambiguity for the compiler. So `len` is OK field name. :)

fexolm: @Quasilyte, thanks for correction. Anyway we need to rename the variable :)

PR #122, opened 2018-05-28
lint: add range-expr-copy checker

### LINKED ISSUE #120: lint: add array range copy checker
For range loops of form `for k, v := range a`, where `a` is an array,
emit a warning if array is big enough to cause measurable impact on performance due to copying.

This code triggers a warning:
```go
var xs [256]byte
/// copy of xs (256 bytes) can be avoided with &xs
for _, x := range xs {
	_ = x
}
```

There are corner cases that must be handled.
For example, if range expression is not addressable, we can't suggest `&a` alternative:
```go
func returnArray() [20]int {
	return [20]int{}
}

func noWarnings() {
	// OK: returned valus is not addressible, can't take address.
	for _, x := range returnArray() {
		_ = x
	}
}
```

The size threshold is needed to avoid bad advices.
Changing small array loop that copies data to `&a` form can cause slowdown on some
machines. Empirically found more-or-less optimal threshold of 64 bytes (6 machine words on x64), but we should probably stick to 96 bytes to avoid neutral changes that give no stable benefits.

Checker suggested by @fexolm.

Quasilyte:
Fixes #120

### DIFF
--- a/lint/range-expr-copy_checker.go
+++ b/lint/range-expr-copy_checker.go
@@ -1,0 +1,52 @@
+package lint
+
+import (
+	"go/ast"
+	"go/types"
+)
+
+// rangeExprCopyCheck detects for statements with range expressions
+// that perform excessive copying (big arrays can cause it).
+//
+// Suggested fix is to use & and doing a range over pointer to array.
+//
+// Rationale: performance.
+func rangeExprCopyCheck(ctx *context) func(*ast.File) {
+	// TODO(quasilyte): there is some annoying code duplication with other
+	// range statement checker. We should consider refactoring if
+	// more checkers that inspect range statements will appear.
+	return wrapStmtChecker(&rangeExprCopyChecker{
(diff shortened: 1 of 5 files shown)


lint/range-expr-copy_checker.go
```diff
@@ -0,0 +1,52 @@
...
+		baseStmtChecker: baseStmtChecker{ctx: ctx},
+	})
+}
+
+type rangeExprCopyChecker struct {
+	baseStmtChecker
+}
+
+func (c *rangeExprCopyChecker) PerFuncInit(fn *ast.FuncDecl) bool {
+	return fn.Body != nil && !c.ctx.IsUnitTestFuncDecl(fn)
```
fexolm: ```c.ctx.IsUnitTestFuncDecl(fn)``` why is it necessary?

Quasilyte: Because many tests use table-driven approach and it leads to kind of false-positive warnings.

PR #126, opened 2018-06-02
add docs for param-duplication

fexolm:


docs/checkers/param-duplication.md
```diff
@@ -0,0 +1,13 @@
+## param-duplication
```
Quasilyte: I think adding this header is redundant as filename identifies checker/rule name uniquely.
And this file alone is not a complete reference, only partial documentation.

Quasilyte:
LGTM.

docs/checkers/comments.md
```diff
@@ -0,0 +1,19 @@
+Detects uninformative comments
```
Quasilyte: Detects comments that aim to silence `go lint` complaints about exported symbol not having a doc-comment.
(Also note that period is missing.)

PR #130, opened 2018-06-02
cmd/kfulint/testdata: should separate positive and negative tests

### LINKED ISSUE #108: cmd/kfulint/testdata: should separate positive and negative tests
Right now all tests are intermixed.
We should separate positive and negative tests for better maintainability.

weeellz:
Fixes #108.


Quasilyte:
I have an idea.
1. Revert to initial state prior to this commit.
2. Change `checker_tests` name to `positive_tests` in `lint_test.go`.
3. Rename all `checker_tests.go` files in `cmd/kfulint/testdata` to `positive_tests.go`.
4. Next step would be to move out negative tests to a separate file.

This way, only `positive_tests.go` supports directives, but all files from the tests directory are checked together, so if any warnings report issues in non-positive tests file, we're going to get "unexpected output" lines.

It's OK to only focus on `switchif` checker for this PR.
It would be easy to separate tests after this PR is merged.

weeellz:
@Quasilyte, how to revert to initial state?

Quasilyte:
@weeellz, there are multiple options, depending on whenever you want to preserve current changes or not.

The simplest way is to create new PR and start from scratch, branching from master.

If you want to continue on this PR, you can do either of these:
a) Checkout master, pull latest changes, copy changed file(s), checkout your branch, paste those files. And you can continue. Next commit will include only wanted changes. When we will do a squash, all unwanted history will be removed.
b) Revert all commits from the beginning of this branch. Then you can continue from scratch. This creates several (or one combined) revert commits.
c) You can do `git reset HEAD~x` where `x` is a number of commits you want to remove. Then you do `git push -f origin weeellz/separate_tests`.

I think @fexolm, can suggest other options. :)

Quasilyte:
The `big-copy` checker was recently renamed to `range-val-copy` checker.

Quasilyte:
Otherwise looks good.

cmd/kfulint/testdata/range-val-copy/positive_tests.go
```diff
@@ -0,0 +1,67 @@
+package checker_test
```
fexolm: Looks like merge issue.

Quasilyte: Why do you think so?

Quasilyte: I'm not sure what caused that.
We can merge this PR and then send another one that fixes it by removing/renaming incorrect things.

PR #132, opened 2018-06-03
cmd/docgen: implement documentation generator

fexolm:
fixes #132


Quasilyte:
Is `cmd/makedocs/makedocs` binary file required?

cmd/makedocs/main.go
```diff
@@ -0,0 +1,65 @@
+package main
+
+import (
+	"html/template"
```
Quasilyte: You should probably use `text/template`.
This way, there shouldn't be any weird escaping problems in the output.

docs/overview.md
```diff
@@ -0,0 +1,370 @@
...
+```go
+a := q.w.e.r.t &#43; 1
+b := q.w.e.r.t &#43; 2
+c := q.w.e.r.t &#43; 3
+v := (a&#43;xs[i&#43;1]) &#43; (b&#43;xs[i&#43;1]) &#43; (c&#43;xs[i&#43;1])
+```
+
+**After**
+```go
+x := xs[i&#43;1]
```
Quasilyte: These escaped sequences are probably there due to `html/template` usage.

cmd/makedocs/main.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	file, err := os.OpenFile(docsPath+"overview.md", os.O_WRONLY|os.O_CREATE, 0600)
+	if err != nil {
+		log.Fatal(err)
+	}
+	err = tmpl.Execute(file, struct {
+		Checkers []checker
+	}{
+		Checkers: checkers,
+	})
```
Quasilyte: You can use `bytes.Buffer` for `tmpl.Execute` and then do `ioutil.WriteFile`.
This way, you don't need to mess around `OpenFile` and it's flags, you don't need to close file either.

cmd/makedocs/main.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+		checkers = append(checkers, checker{
+			Name:         r.Name(),
+			Experimental: r.Experimental(),
+			Description:  desc,
+		})
+	}
+	if err != nil {
+		log.Fatal(err)
+	}
```
Quasilyte: What error does it check?
Seems like these 3 lines are not needed here.

cmd/makedocs/main.go
```diff
@@ -0,0 +1,63 @@
...
+	buf := bytes.Buffer{}
+	err = tmpl.Execute(&buf, struct {
+		Checkers []checker
+	}{
+		Checkers: checkers,
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+	ioutil.WriteFile(docsPath+"overview.md", buf.Bytes(), 0600)
```
Quasilyte:
```diff
- ioutil.WriteFile(docsPath+"overview.md", buf.Bytes(), 0600)
+ if err := ioutil.WriteFile(docsPath+"overview.md", buf.Bytes(), 0600); err != nil {
+ 	log.Fatalf("write output file: %v", err)
+ }
```

PR #139, opened 2018-06-04
cmd/kfulint/testdata: separate positive and negative tests

### LINKED ISSUE #108: cmd/kfulint/testdata: should separate positive and negative tests

weeellz:
Fixes #108.


cmd/kfulint/testdata/big-copy/negative_tests.go
```diff
@@ -0,0 +1,41 @@
+package checker_test
```
Quasilyte: We renamed `big-copy` checker.
It should be called `range-val-copy` now.

Quasilyte:
With exception of renamed checker, looks good. Thank you!

cmd/kfulint/testdata/big-copy/negative_tests.go
weeellz: Oops, fixed.

PR #141, opened 2018-06-04
docs: add step-by-step guide that describes how to add new checker

### LINKED ISSUE #116: docs: add step-by-step guide that describes how to add new checker
In short, the process is like:

1. Come up with checker idea. Concentrate on *what it checks*.
2. Select one of the base checkers kind (example: expr checker, stmt checker, etc.).
3. Define checker type and constructor function.
4. Add entry to checkers list in `lint.go`. It could be a good idea to mark recently added checker as `experimental`.
5. Add test directory that is named after the checker in `cmd/kfulint/testdata`.
6. Add `checker_tests.go` file in that directory and add some positive and negative tests there.
7. Run tests. They must fail as your checker does not check anything yet. Tests can be run with 
`go test -v github.com/PieselBois/kfulint/cmd/kfulint`.
8. Implement checker itself. Make tests pass.

weeellz:
Fixes #116.


weeellz:
@tengufromsky, PTAL.

PR #146, opened 2018-06-05
lint: do not use reference types in function parameters/results

cristaloleg:
Warn users when passing pointer to chan/map
```
func f(ch *chan string) (ch2 *chan *int) {
	return nil
}
```

### DIFF
--- a/lint/ptr-to-ref-type-param_checker.go
+++ b/lint/ptr-to-ref-type-param_checker.go
@@ -16,9 +16,7 @@ type ptrToRefTypeParamChecker struct {
 }
 
 func (c *ptrToRefTypeParamChecker) CheckFuncDecl(fn *ast.FuncDecl) {
-	if fn.Type.Params != nil {
-		c.checkParams(fn.Type.Params.List)
-	}
+	c.checkParams(fn.Type.Params.List)
 	if fn.Type.Results != nil {
 		c.checkParams(fn.Type.Results.List)
 	}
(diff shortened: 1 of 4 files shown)


lint/ptr-to-ref-type-param_checker.go
```diff
@@ -0,0 +1,57 @@
...
+		baseFuncDeclChecker: baseFuncDeclChecker{ctx: ctx},
+	})
+}
+
+type ptrToRefTypeParamChecker struct {
+	baseFuncDeclChecker
+}
+
+func (c *ptrToRefTypeParamChecker) CheckFuncDecl(fn *ast.FuncDecl) {
+	if fn.Type.Params != nil {
```
Quasilyte: `Params` never nil. Documented in `go/ast`.

lint/ptr-to-ref-type-param_checker.go
```diff
@@ -0,0 +1,57 @@
...
+func (c *ptrToRefTypeParamChecker) checkParams(params []*ast.Field) {
+	for _, param := range params {
+		ptr, ok := c.ctx.TypesInfo.TypeOf(param.Type).(*types.Pointer)
+		if !ok {
+			continue
+		}
+
+		if isRefType(ptr.Elem().Underlying()) {
+			if len(param.Names) == 0 {
+				c.ctx.Warn(param, "consider to make returning type of non-pointer type")
```
Quasilyte: For input params, this will print confusing message about returning type.
It's OK for the first version though, just making an observation.

cristaloleg: `consider to make type of non-pointer type` ?

Quasilyte: That will work, yes.

It would be even better if type would be printer in error message.
Type string can be created with `nodeString(c.ctx.FileSet, param.Type)`.

PR #150, opened 2018-06-05
lint: minor fixes to #146

Quasilyte:
- Make checker type name (and filename) consistent with rule name
- Make referential type predicate function checker method.
  Not every checker may want to define slice as referential type.
- s/ch/slice/ for parameter which is slice (to avoid confusion)

Also, I've made nickname typo in branch name. :3

### DIFF
--- a/lint/lint.go
+++ b/lint/lint.go
@@ -32,7 +32,7 @@ var checkFunctions = map[string]struct {
 	"builtin-shadow":    {new: builtinShadowCheck},
 	"range-expr-copy":   {new: rangeExprCopyCheck},
 	"stddef":            {new: stddefCheck},
-	"ptr-to-ref-param":  {new: ptrToRefTypeParamCheck},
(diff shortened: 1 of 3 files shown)


Quasilyte:
CC @cristaloleg

lint/lint.go
```diff
@@ -32,7 +32,7 @@ var checkFunctions = map[string]struct {
 	"builtin-shadow":    {new: builtinShadowCheck},
 	"range-expr-copy":   {new: rangeExprCopyCheck},
 	"stddef":            {new: stddefCheck},
-	"ptr-to-ref-param":  {new: ptrToRefTypeParamCheck},
+	"ptr-to-ref-param":  {new: ptrToRefParamCheck},
```
Quasilyte: The other way is to make rule name `ptr-to-ref-type-param`, but I would prefer shorter (current) form.

Issue #153, opened 2018-06-06
all: complete transfer to new org and repo


Quasilyte:
Update docs and all other things.

☑ fix import paths
☑ ensure ci-pr works
☑ ensure ci-push works
☑ update readme (including URLs in badges)

fexolm:
@Quasilyte, done?

Quasilyte:
I don't know. I haven't performed any good investigation on what else needs to be done.
If you have additions, edit original (first) message with empty checkbox. Thank you.

PR #157, opened 2018-06-06
readme: updates due to project name change

Quasilyte:
Also changed go-critic binary name to gocritic as
dash is sometimes used for sub commands in linux systems.
"gometalinter" also does not use dash in it's cmd name.

Updates #153


Quasilyte:
@weeellz, could you do a review?

PR #160, opened 2018-06-07
lint: quickfix implementation [doesn't work for now]

fexolm:
This implementation doesn't work. File doesn't change. Anyone knows why? 
It looks like i edit copy of ast.FuncDecl.

### DIFF
--- a/cmd/gocritic/main.go
+++ b/cmd/gocritic/main.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"bytes"
 	"flag"
 	"fmt"
 	"go/ast"
 	"go/parser"
+	"go/printer"
 	"go/types"
 	"log"
 	"os"
@@ -27,6 +29,8 @@ type linter struct {
 
 	packages        []string
 	enabledCheckers []string
+
+	quickFix bool
 }
 
 func main() {
@@ -57,6 +61,7 @@ func parseArgv(l *linter) {
 	}
 
 	enable := flag.String("enable", "all", "comma-separated list of enabled checkers")
+	l.quickFix = *flag.Bool("quickfix", false, "enables quickfix mode")
 
 	flag.Parse()
 
@@ -104,7 +109,11 @@ func (l *linter) LoadProgram() {
 	l.ctx = &lint.Context{
 		FileSet:   prog.Fset,
 		SizesInfo: sizes,
+		QuickFix:  true,
 	}
+
+	log.Print(l.ctx.QuickFix)
+
 }
 
 func (l *linter) InitCheckers() {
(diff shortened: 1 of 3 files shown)


lint/param-duplication_checker.go
```diff
@@ -19,7 +19,11 @@ type paramDuplicationChecker struct {
 func (c *paramDuplicationChecker) CheckFuncDecl(decl *ast.FuncDecl) {
 	typ := c.optimizeFuncType(decl.Type)
 	if !astcmp.EqualExpr(typ, decl.Type) {
-		c.warn(decl.Type, typ)
+		if c.ctx.QuickFix {
+			decl.Type = typ
```
fexolm: it seems like that line does nothing.

Quasilyte:
I'll take a look today or tomorrow.

cmd/gocritic/main.go
```diff
@@ -57,6 +61,7 @@ func parseArgv(l *linter) {
 	}
 
 	enable := flag.String("enable", "all", "comma-separated list of enabled checkers")
+	l.quickFix = *flag.Bool("quickfix", false, "enables quickfix mode")
```
Quasilyte:
```diff
- l.quickFix = *flag.Bool("quickfix", false, "enables quickfix mode")
+ flag.Var(&l.quickFix, "quickfix", false, "enables quickfix mode")

Comment (Quasilyte) [reply cmd/gocritic/main.go]: Otherwise quickFix will be always `false`.

File: lint/lint.go [hunk 3]
```diff
@@ -97,8 +98,17 @@ type Checker struct {
...
+		c.ctx.FileMutex.Lock()
+		c.check(f)
+		c.ctx.FileMutex.Unlock()
+	} else {
+		c.ctx.FileMutex.RLock()
+		c.ctx.warnings = c.ctx.warnings[:0]
+		c.check(f)
+		c.ctx.FileMutex.RUnlock()
+
+	}
```
Quasilyte: If I understand it correctly, this kind of locking will make our code 100% linear (serialized).
First checker will acquire lock and all others will have to wait until it finishes.

cmd/gocritic/main.go
```diff
@@ -176,11 +185,20 @@ func (l *linter) checkFile(f *ast.File) {
...
 			}()
 
-			for _, warn := range c.Check(f) {
-				pos := l.ctx.FileSet.Position(warn.Node.Pos())
-				log.Printf("%s: %s: %v\n", pos, c.Rule, warn.Text)
+			if !l.ctx.QuickFix {
+				for _, warn := range c.Check(f) {
+					pos := l.ctx.FileSet.Position(warn.Node.Pos())
+					log.Printf("%s: %s: %v\n", pos, c.Rule, warn.Text)
+				}
```
Quasilyte: You don't call `c.Check` if `l.ctx.QuickFix` is true.
So, with quickfix, no checkers are executed.
If you will actually execute them, updated AST will be printed, as expected (I've verified that).

lint/lint.go
Quasilyte: Correction: not entirely serialized, but it will lead to N parallelism instead of M*N.
Usually, locking should be done before and after and operation (as you do it there), but for the minimal amount of time required. In this program, you want to do lock+unlock around code that modifies the tree:
```go
if c.ctx.QuickFix {
	decl.Type = typ
}
```

cmd/gocritic/main.go
fexolm: Oh, god... Stupid mistake :D Thanks for finding it.

fexolm:
I have recieved the answer. So, I will close pr. Thank you @Quasilyte!

PR #163, opened 2018-06-07
lint: add flag-deref checker

### LINKED ISSUE #162: lint: add flag-deref checker
Detects invalid usages of `flag` package.

**Before:**
```go
a := *flag.Bool("a", false, "a bool flag")
```

**After:**
```go
var a bool
flag.BoolVar(&a, false, "a bool flag")
```

Quasilyte:
Fixes #162

### DIFF
--- a/docs/checkers/flag-deref.md
+++ b/docs/checkers/flag-deref.md
@@ -1,0 +1,13 @@
+Detects immediate dereferencing of `flag` package pointers.
+Suggests using `XxxVar` functions to achieve desired effect.
+
+**Before:**
+```go
+b := *flag.Bool("b", false, "b docs")
+```
+
+**After:**
+```go
+var b bool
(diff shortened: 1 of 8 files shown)


docs/checkers/flag-deref.md
```diff
@@ -0,0 +1,13 @@
+Detects immediate dereferencing of `flag` package pointers.
+Suggests using `XxxVar` functions to achieve desired effect.
```
cristaloleg: Can you add a bit more detailed explanation from issue? 😉

Quasilyte: PTAL.

Quasilyte: `Dereferencing returned pointers will lead to hard to find errors where flag values are not updated after flag.Parse().`

Issue #21, opened 2018-05-08
lint: add normalization checks


Quasilyte:
Don't know how to categorize changes, but we can detect several excessive AST elements.

```go
switch true {/* body */}
// =>
switch {/* body */}

// This case is handled by gosimple:
for true {/* body */}
// =>
for {/* body */}
```
There could be more.
We must implement only new checks that are not present in, say, [gosimple](https://github.com/dominikh/go-tools/tree/master/cmd/gosimple).

It can also be a good idea to report them in gosimple issue tracker.
It's strange that it handles `for` statement, but not `switch`.
Maybe there is a rationale for that.

PR #166, opened 2018-06-08
lint: add switch-true checker

Quasilyte:
Does trigger at least 2 times on GOROOT:
  math/sinh.go:46: switch true {
  net/ip.go:225: switch true {

Updates #21


Quasilyte:
I've done some refactoring so can we merge this and leave fmt.Sprintf("") checker for "first contributors"?

PR #173, opened 2018-06-09
cmd/makedocs: render experimental and other checkers separately

### LINKED ISSUE #170: cmd/makedocs: render experimental and other checkers separately
It's better to render 2 tables: one for experimental and second for every other checker.
This way, we don't need to have second column that is not very informative and "false" in 90+% of cases.

```
Checkers:
a
b
c
d

Experimental checkers:
e
f
```

Instead of:
```
Checkers:
a false
b false
c false
d false
e true
f true
```

fexolm:
fixes #170


docs/overview.md
```diff
@@ -4,95 +4,83 @@
...
+      </tr>
+      <tr>
+        <td><a href="#unexported-call-ref">unexported-call</a></td>
+      </tr>
+      <tr>
+        <td><a href="#unslice-ref">unslice</a></td>
+      </tr>
+</table>
+
+**Experimental:**
```
Quasilyte: "Experimental checkers:"?

Issue #99, opened 2018-05-19
cmd/lintwalk: same package execution multiple times


weeellz:
E.g. when doing `lintwalk -src-root=$GOROOT/src -enable='switchif'`
Output:
```/home/ravil/Go/go1.10/src/net/error_test.go:77:2: switchif: should rewrite switch statement to if statement
/home/ravil/Go/go1.10/src/net/error_test.go:414:2: switchif: should rewrite switch statement to if statement
/home/ravil/Go/go1.10/src/net/error_test.go:431:2: switchif: should rewrite switch statement to if statement
/home/ravil/Go/go1.10/src/net/error_test.go:77:2: switchif: should rewrite switch statement to if statement
/home/ravil/Go/go1.10/src/net/error_test.go:414:2: switchif: should rewrite switch statement to if statement
/home/ravil/Go/go1.10/src/net/error_test.go:431:2: switchif: should rewrite switch statement to if statement
```
And counting...

Quasilyte:
Will fix as a part of #158.

Quasilyte:
Can't reproduce using master branch version of linter.
```
net/error_test.go:79:2: switchif: should rewrite switch statement to if statement
net/error_test.go:103:2: switchif: should rewrite switch statement to if statement
net/error_test.go:140:2: switchif: should rewrite switch statement to if statement
net/error_test.go:187:2: switchif: should rewrite switch statement to if statement
net/error_test.go:216:2: switchif: should rewrite switch statement to if statement
net/error_test.go:289:2: switchif: should rewrite switch statement to if statement
net/error_test.go:349:2: switchif: should rewrite switch statement to if statement
net/error_test.go:378:2: switchif: should rewrite switch statement to if statement
net/error_test.go:416:2: switchif: should rewrite switch statement to if statement
net/error_test.go:433:2: switchif: should rewrite switch statement to if statement
net/error_test.go:438:2: switchif: should rewrite switch statement to if statement
net/error_test.go:459:2: switchif: should rewrite switch statement to if statement
net/error_test.go:480:2: switchif: should rewrite switch statement to if statement
net/error_test.go:511:2: switchif: should rewrite switch statement to if statement
net/error_test.go:533:2: switchif: should rewrite switch statement to if statement
net/error_test.go:543:2: switchif: should rewrite switch statement to if statement
net/error_test.go:607:2: switchif: should rewrite switch statement to if statement
net/error_test.go:629:2: switchif: should rewrite switch statement to if statement
[...]

PR #176, opened 2018-06-09
cmd: apply sub-command design to gocritic

Quasilyte:
gocritic is user API for the linter.
It has several sub-commands that should be stable,
but implementation can change in any way we want.

Currently there are only 2 sub commands:
- check-package (previously gocritic)
- check-project (previously lintwalk)

Example:
```
# New command:
	gocritic check-package a b
# Old command:
	gorcitic a b
```

### DIFF
--- a/cmd/criticize/main.go
+++ b/cmd/criticize/main.go
@@ -1,0 +1,185 @@
+package criticize
+
+import (
+	"flag"
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/types"
+	"log"
+	"os"
+	"runtime"
+	"strings"
+	"sync"
+
+	"github.com/go-critic/go-critic/lint"
+	"golang.org/x/tools/go/loader"
+)
+
+type linter struct {
+	ctx *lint.Context
+
+	prog *loader.Program
+
+	checkers []*lint.Checker
+
+	// Command line flags:
+
+	packages        []string
+	enabledCheckers []string
+}
+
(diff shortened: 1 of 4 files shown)


Quasilyte:
I've choosed `cmd/go` approach.

cmd/criticize/main.go
```diff
@@ -0,0 +1,185 @@
...
+
+// parseArgv processes command-line arguments and fills ctx argument with them.
+// Terminates program on error.
+func parseArgv(l *linter) {
+	flag.Usage = func() {
+		fmt.Fprintln(os.Stderr, "usage: gocritic [flags] [package]")
+		flag.PrintDefaults()
+	}
+
+	enable := flag.String("enable", "all", "comma-separated list of enabled checkers")
```
cristaloleg: `flag.StringVar` ? 😉

Quasilyte:
Note that his does not fix #99.
This change only adds additional layer above existing commands.
I'll address that bug in near time.

Quasilyte:
> but maybe rename lintwalk and criticize into check-package and check-project?

Package names with slash require import aliases.
We may rename both packages, but that's the point: we can rename them, or implement `check-package` with multiple packages and choose one using some other context.

This is why `criticize` does know that it's called `check-package` in `gocritic`. 

Decoupling of gods. 80lvl.

fexolm:
Ok then :)

PR #184, opened 2018-06-10
lint: remove checker names map

fexolm:


lint/append-combine_checker.go
```diff
@@ -7,16 +7,23 @@ import (
...
-func appendCombineCheck(ctx *context) func(*ast.File) {
-	return wrapBlockChecker(&appendCombineChecker{
-		baseStmtListChecker: baseStmtListChecker{ctx: ctx},
-	})
+func init() {
+	addChecker(appendCombineChecker{})
 }
 
 type appendCombineChecker struct {
+	stableChecker
```
Quasilyte: Hmm, the idea is interesting, but what about this point of this kind of solution:
The "experimental" or "stable" are not properties of checker themselves of their traits, but rather
an info added when rule-checker relation is created.
In terms of code:
```go
// For stable checkers
func init() {
  addChecker(...)
}
// For unstable checkers
func init() {
  addExperimentalChecker(...)
}
```

Although this idea does not scale for other attributes (but we have none of them yet)..
What do you think about making `checkFunction` an struct with `experimental bool` field. For most checkers we don't need explicitly set it because default value is `false`.
The constructor can be a field of type `func(ctx *context) func(*ast.File)`.

fexolm: I have fixed `New`, but I am not shure what idea with experimantal is better.

Quasilyte: What do you think about structs?
```go
func init() {
  constructor := func (ctx *context) func(*ast.File) {
    return wrapBlockChecker(&appendCombineChecker{
      baseStmtListChecker: baseStmtListChecker{ctx: ctx},
    })
  }
  addRule(ruleInfo{new: constructor})
}
```
The main advantage is that you don't need:
- checkFunc interface
- implement anything on checker except checking
- specify experimental=false, only if it's true (because of default false value)
- this scales better when there are more optional attributes that have default value for most frequent cases

Quasilyte: ruleInfo is:
```go
type ruleInfo struct {
	experimental bool
	new func(*context) func(*ast.File)
}
```

fexolm: I don't understand how it will get type name in that implementation

Quasilyte: Hmm. Please interpret `ruleInfo` as `checkFunc`. Our terminology is a bit messy here. We may cleanup it later. Exact names are not that relevant right now.

Quasilyte:
I don't like interface thing and New method in particular, but can't propose better options right now, so you can merge.

PR #194, opened 2018-06-11
cmd/lintwalk: support sub-projects and arbitrary src roots 

### LINKED ISSUE #192: cmd/lintwalk: support sub-projects and arbitrary src roots
Right now `lintwalk` can only handle topmost projects, that is, `GOPATH/src` or `GOROOT/src`.
It would be more useful with arbitrary source dirs without that restriction.

It may require appending `GOPATH` -> `src dir` path to get valid full imports for `check-package`.

fexolm:
fixes #192

### DIFF
--- a/cmd/lintwalk/main.go
+++ b/cmd/lintwalk/main.go
@@ -2,18 +2,43 @@ package lintwalk
 
 import (
 	"flag"
+	"go/build"
+	"go/parser"
+	"go/token"
 	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"runtime"
+	"strings"
 )
+
+func packagePath() []string {
+	res := []string{runtime.GOROOT(), build.Default.GOPATH}
+
+	if p, ok := os.LookupEnv("GOPATH"); ok {
+		res = append(res, strings.Split(p, ":")...)
+	}
+	return res
+}
+
+func getPackagePrefix(dir string) string {
+	for _, p := range packagePath() {
+		if filepath.HasPrefix(dir, p) {
+			if res, err := filepath.Rel(filepath.Join(p, "src"), dir); err == nil {
+				return res
+			}
+		}
+	}
+	return ""
+}
 
 // Main implements gocritic sub-command entry point.
 func Main() {
 	enable := flag.String("enable", "all",
 		`forwarded to linter "as is"`)
-	exclude := flag.String("exclude", "testdata/|vendor/|builtin/",
+	exclude := flag.String("exclude", "(.*test.*)|(.*vendor.*)|(.*builtin.*)",
 		`regexp used to skip package names`)
 
 	flag.Parse()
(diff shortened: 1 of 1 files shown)


cmd/lintwalk/main.go
```diff
@@ -27,39 +52,61 @@ func Main() {
...
+		log.Fatal(err)
+	}
+
 	excludeRE, err := regexp.Compile(*exclude)
+
+	testRE := regexp.MustCompile(".*_test")
+
 	if err != nil {
-		log.Fatalf("bad -exclude pattern: %v", err)
+		log.Fatal(err)
```
Quasilyte: Hmm, I wonder why you think it's better to exclude error context from `log.Fatal` message.

cmd/lintwalk/main.go
```diff
@@ -2,18 +2,43 @@ package lintwalk
...
+	}
+	return ""
+}
+
 // Main implements gocritic sub-command entry point.
 func Main() {
 	enable := flag.String("enable", "all",
 		`forwarded to linter "as is"`)
-	exclude := flag.String("exclude", "testdata/|vendor/|builtin/",
+	exclude := flag.String("exclude", "(.*test.*)|(.*vendor.*)|(.*builtin.*)",
```
Quasilyte: This will skip more than required.
Older pattern matched `/test/` somewhere in the path, new regexp will match things like `testify` or `contest` (not related to tests at all). Is that intentional?

cmd/lintwalk/main.go
```diff
@@ -27,39 +52,61 @@ func Main() {
...
+		if !info.IsDir() || excludeRE.MatchString(p) {
 			return nil
 		}
-		if excludeRE.MatchString(path) {
-			return nil
+		p = filepath.Dir(p)
+		p = getPackagePrefix(p)
+
+		for _, pk := range pkgs {
+			if testRE.MatchString(pk.Name) {
```
Quasilyte: What about using `strings.HasSuffix(pk.Name, "_test.go")`?

Quasilyte: HasSuffix is Go "EndsWith" function.

Quasilyte:
I'm concerned that lintwalk does not check external test now (packages with `_test` suffix).
It did check them before.
Are there plans to return this behavior? Otherwise it looks like introducing one feature at the cost of other feature.

Quasilyte:
@cristaloleg, can you lend a hand here and think about what can be improved?
If we are going to skip `_test` package this should be either documented or fixed in next patches.

My opinion is that tests should be just as readable and correct as normal code.

cmd/lintwalk/main.go
```diff
@@ -2,18 +2,43 @@ package lintwalk
...
+	"runtime"
+	"strings"
 )
 
+func packagePath() []string {
+	res := []string{runtime.GOROOT(), build.Default.GOPATH}
+
+	if p, ok := os.LookupEnv("GOPATH"); ok {
+		res = append(res, strings.Split(p, ":")...)
+	}
```
Quasilyte: I like the intention of handling more corner cases, but ~ no one uses multiple GOPATHs since most Go tools and linters can't handle it. Text editors can't do it either (not even sure IDEs can).
This feature can even be deprecated in near future (especially since vgo is approaching).

Also note that about 70% of Go programmers do not know about `:` and GOPATH lists, so you should comment what you're doing in that loop and why it's required.
The other option is to remove it and ignore it's existence.

fexolm:
Fixed. Now we don't ignore anything except "exclude"list.

Quasilyte:
@fexolm , sounds cool!
I'll take a look in near hour.

PR #195, opened 2018-06-11
cmd/gocritic: exit with non-zero status on errors

### LINKED ISSUE #93: cmd/kfulint: exit with non-zero status if there were any linting errors
Should do `os.Exit(1)` if any warnings have triggered.
Most likely we can add some field into `linter` type that is set to true (or exit code value) when warnings are being printed. There can be other approaches to this, but we should avoid global variables.

Quasilyte:
Fixes #93


Quasilyte:
I see. Test runner interprets non-zero exit status as an error.

Quasilyte:
Now it works. Awaiting for re-approval.

PR #197, opened 2018-06-11
lint: teach stddef to avoid pkgs that define suggested consts

### LINKED ISSUE #181: lint/stddef: triggers on itself
That's weird.
Hope it doesn't trigger on math or bits package.
Need to investigate potential fixes.
Maybe package name filters are good enough, but then how to avoid skipping `lint` package?

Quasilyte:
Fixes #181


Quasilyte:
Now it doesn't suggest to use math constants in math package, or time constants in time package.

Quasilyte:
Now it doesn't trigger on itself either.

Issue #200, opened 2018-06-11
lint: consider using astp package


Quasilyte:
I'll reference this issue from commits that may benefit from [astp](https://github.com/cristaloleg/astp) package.

CC @cristaloleg

Quasilyte:
```diff
- paren := findNode(expr, func(x ast.Node) bool {
-	_, ok := x.(*ast.ParenExpr)
-	return ok
- })
+ paren := findNode(expr, astp.IsParenExpr)
```

PR #201, opened 2018-06-12
lint: improve type parenthesis checker warning messages

### LINKED ISSUE #5: lint,cmd/kfulint: record types info and parsed ast files inside context
Currently, `ast.ParseDir` results are ignored.
It returns multiple packages, usually it's target package (same name as containing directory name) only,
but sometimes there is also testing package with `_test` suffix.

Proposed plan:
1. Collect parsed packages.
2. Iterate over these packages, init context with current package files and types info.
3. Run checkers over that package with properly initialized context.
4. Repeat until there are no packages left.

**Note**: we may reject multi-package folders for now.
In other words: it's OK to exit if `len(packages) != 1` and that package name does not
match target package name. "proper" multi-package handling may be added later.
This can simplify initial implementation significantly.

`lint.Context` needs additional fields.
It definitely needs `*types.Info` and `[]*ast.File`. Maybe something else.

Quasilyte:
Fixes #5
Updates #200


Quasilyte:
@fexolm, I'll move it to separate repo to then. :)
Will send updated PR later.

Quasilyte:
I'm almost ready.
https://github.com/go-toolsmith/astcopy
Expect another update this evening.

Quasilyte:
Don't look into `astcopy.go` code. It's full of pain and misery.

PR #204, opened 2018-06-12
readme: add dependencies list

Quasilyte:
Gives [go-toolsmith](https://github.com/go-toolsmith/) visibility. :)


fexolm:
@Quasilyte, you sure, that we need that in readme? for user it looks like, "oh, i have to download many dependencies..." Maybe we could add link in "How to add new checker" section? Like "see also go-toolsmith"?

Quasilyte:
OK. Convinced.

PR #217, opened 2018-06-15
lint: use astfmt instead of nodeString

Quasilyte:


Quasilyte:
I'll wait until @fexolm has his chance to notice this.

PR #221, opened 2018-06-16
lint: unnamedResults checker

### LINKED ISSUE #207: lint: check unnamed returns with similar type
before:
```go
func f() (float64, float64)
```
after:
```go
func f() (x, y float64)
```
This will make function definition much easier to read & understand.
We might also be interested to warn users about unnamed params/returns in method declaration too.

cristaloleg:
Fixes #207


lint/unnamedResult_checker.go
```diff
@@ -0,0 +1,61 @@
...
+	c.ctx.Warn(n, "consider to give name to results")
+}
+
+func (c *unnamedResultChecker) isBoolOrError(expr ast.Expr) bool {
+	switch typ := c.ctx.TypesInfo.TypeOf(expr).(type) {
+	case *types.Named:
+		return typ.Obj().Name() == "error"
+
+	case *types.Basic:
+		return typ.Name() == "bool"
```
Quasilyte: Name comparison will work, but for the reference, you can also do this:
```go
return typ.Kind() == types.Bool
```

cmd/gocritic/testdata/unnamedResult/negative_tests.go
```diff
@@ -0,0 +1,29 @@
+package checker_test
+
```
Quasilyte: Negative tests also need a few declarations with named results with 2 and more results that would emit warning
if there were unnamed.

Quasilyte:
LGTM when at least 1 negative test with named results is added.

lint/unnamedResult_checker.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (c *unnamedResultChecker) warn(n ast.Node) {
+	c.ctx.Warn(n, "consider to give name to results")
+}
+
+// resultsNum will return number of results.
+// If they're unnamed than we meed to return number of fields.
+// Else we need to sum all amount of names in each field.
+func (c *unnamedResultChecker) resultsNum(fields []*ast.Field) int {
```
Quasilyte: Almost looks like `FieldList.NumFields`.
Can we use that instead or I'm missing something?

Quasilyte:
Thank you!

PR #223, opened 2018-06-16
docs: slightly reword unnamedResult docs

Quasilyte:


docs/overview.md
```diff
@@ -581,6 +582,8 @@ Gives false-positives for:
 
 
 
+
+
```
cristaloleg: I think we should fix `makedocs` tool :)

Quasilyte: Maybe. I don't know why it did that..
Needs investigation and dedicated issue.

Quasilyte: Could you file an issue?

cristaloleg: yep

PR #229, opened 2018-06-16
all: add syntaxonly attribute for rules

### LINKED ISSUE #226: lint: add `astonly` attribute for checkers
This attribute tells that checker does not need type info to function properly.

`astonly` checkers should not access `Context.TypeInfo`.

Note that default value is `false` and even if checker is effectively `astonly`, it can be unmarked if there are plans to use type info later.

Rationale: can be useful for golangci-lint integration; parsing is much-much faster than real program loading or entire import tree type checking (orders of magnitude faster).

Refs #222

Quasilyte:
Also updated makedocs to output Kind column.

Fixes #226

### DIFF
--- a/cmd/makedocs/main.go
+++ b/cmd/makedocs/main.go
@@ -19,6 +19,7 @@ type checker struct {
 	Name         string
 	Description  string
 	Experimental bool
+	Kind         string
 }
 
 func main() {
@@ -35,6 +36,7 @@ func main() {
 		checkers = append(checkers, checker{
 			Name:         r.Name(),
 			Experimental: r.Experimental(),
+			Kind:         kindString(r),
 			Description:  desc,
 		})
 	}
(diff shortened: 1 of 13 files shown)


Quasilyte:
We may need some refactoring soon.
I'll add new attribute without refactoring for now to make reviewing easier.

lint/lint.go
```diff
@@ -67,6 +88,11 @@ func NewChecker(rule *Rule, ctx *Context) *Checker {
 	if rule == nil {
 		panic("nil rule given")
 	}
+	// TODO(quasilyte): it would be great to have SyntaxOnly
```
cristaloleg: Sounds as an issue. Can you fill it?

cmd/makedocs/main.go
```diff
@@ -56,3 +58,10 @@ func getDesc(name string) (string, error) {
 	b, err := ioutil.ReadFile(checkersPath + name + ".md")
 	return string(b), err
 }
+
+func kindString(r *lint.Rule) string {
```
cristaloleg: s/`kindString`/`ruleKind`?

lint/lint.go
Quasilyte: Sure.

cmd/makedocs/main.go
Quasilyte: What about `ruleKindString`?
I think this "Kind" string may include more attributes later.
It's like human-readable short overview of a set of rule attributes.
(Because it's dumb to fill columns with true/false..)

PR #234, opened 2018-06-16
lint: add filename to context

### LINKED ISSUE #233: lint: add filename to context

cristaloleg:
Fixes #233

### DIFF
--- a/cmd/criticize/main.go
+++ b/cmd/criticize/main.go
@@ -159,6 +159,7 @@ func (l *linter) CheckPackage(pkgPath string) {
 }
 
 func (l *linter) getFilename(f *ast.File) string {
+	// see https://github.com/golang/go/issues/24498
 	fname := l.prog.Fset.Position(f.Pos()).String() // ex: /usr/go/src/pkg/main.go:1:1
 	fname = filepath.Base(fname)                    // ex: main.go:1:1
 	return fname[:len(fname)-4]                     // ex: main.go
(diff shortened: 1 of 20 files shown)


cmd/criticize/main.go
```diff
@@ -153,10 +154,16 @@ func (l *linter) CheckPackage(pkgPath string) {
 	l.ctx.TypesInfo = &pkgInfo.Info
 	l.ctx.Package = pkgInfo.Pkg
 	for _, f := range pkgInfo.Files {
-		l.checkFile(f)
+		l.checkFile(l.getFilename(f), f)
 	}
 }
 
+func (l *linter) getFilename(f *ast.File) string {
+	fname := l.prog.Fset.Position(f.Pos()).String() // ex: /usr/go/src/pkg/main.go:1:1
```
Quasilyte: Could you reference https://github.com/golang/go/issues/24498 in a comment inside `getFilename` function?

cmd/criticize/main.go
```diff
@@ -153,10 +154,16 @@ func (l *linter) CheckPackage(pkgPath string) {
 	l.ctx.TypesInfo = &pkgInfo.Info
 	l.ctx.Package = pkgInfo.Pkg
```
Quasilyte: We can add `Filename` to `lint.Context` instead of `lint.context` and assign to it here.
It's read-only info that can be stored there.

Or did you had something in mind in favor of checker-local context?

PR #235, opened 2018-06-16
lint: warn about unused params

### LINKED ISSUE #208: lint: consider to skip(`_`) unused parameter
Name: `unusedParams`
before:
```
func f(a int, b float64) // b isn't used
```
after:
```
func f(a int, _ float64) // everything is cool
```

cristaloleg:
Fixes #208


cmd/gocritic/testdata/unusedParams/negative_tests.go
```diff
@@ -0,0 +1,12 @@
+package checker_test
+
+func (f *foo) g1(x int, _ float64) {
```
Quasilyte: I think it's intentional that receiver parameter not checked.
Makes sense.

But we should document that explicitly, because it's possible to do:
```go
func (*foo) g1(x int, _ float64) {}
```

lint/unusedParams_checker.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"fmt"
+	"go/ast"
+)
+
+func init() {
+	addChecker(unusedParamsChecker{}, &ruleInfo{})
+}
+
+type unusedParamsChecker struct {
```
Quasilyte: Can we call it `unusedParam` (without "s")?

lint/unusedParams_checker.go
```diff
@@ -0,0 +1,55 @@
...
+	for id := range c.ctx.TypesInfo.Uses {
+		objUsed[id.Obj] = struct{}{}
+	}
+
+	for _, p := range params.List {
+		if len(p.Names) == 0 {
+			c.warnUnnamed(p)
+			break
+		}
+		for _, name := range p.Names {
```
Quasilyte: Hint: calling `p.Names` element as `id` (ident) makes `name.Name`=>`id.Name`, which is somewhat more pleasant.
Note that `Names` are `[]*ast.Ident`, so it does make sense.
(Optional.)

lint/unusedParams_checker.go
```diff
@@ -0,0 +1,55 @@
...
+			}
+			if _, ok := objUsed[name.Obj]; !ok {
+				c.warn(name, name.Name)
+			}
+		}
+	}
+}
+
+func (c *unusedParamsChecker) warn(n ast.Node, param string) {
+	c.ctx.Warn(n, fmt.Sprintf("parameter `%s` isn't used, consider to name it as `_`", param))
```
Quasilyte: Hmm.. Warn accepts format string. It's possible to omit sprintf:
```diff
- c.ctx.Warn(n, fmt.Sprintf("parameter `%s` isn't used, consider to name it as `_`", param))
+ c.ctx.Warn(n, "parameter `%s` isn't used, consider to name it as `_`", param)
```
Additional note: you can print `*ast.Ident` as `%s` directly, thanks to `astfmt`. :)
So, you could accept `*ast.Ident` instead of `string`.
This is up to taste, just a reminder.

lint/unusedParams_checker.go
cristaloleg: sure

lint/unusedParams_checker.go
cristaloleg: omg, didn't now about formatting. I was expecting `Warnf` for that :D

lint/unusedParams_checker.go
cristaloleg: ok

cmd/gocritic/testdata/unusedParams/negative_tests.go
cristaloleg: Good point

lint/unusedParam_checker.go
```diff
@@ -0,0 +1,54 @@
...
+func (c *unusedParamChecker) CheckFuncDecl(decl *ast.FuncDecl) {
+	params := decl.Type.Params
+	if params == nil || params.NumFields() == 0 {
+		return
+	}
+
+	objUsed := make(map[*ast.Object]struct{})
+	for id := range c.ctx.TypesInfo.Uses {
+		objUsed[id.Obj] = struct{}{}
+	}
```
Quasilyte: My main concern is that this checker does whole `TypesInfo.Uses` copy for every declaration.
For stdlib package `reflect`, len(TypesInfo.Uses) is 9092 and checker is invoked on 195 functions there.
1772940 map assignments for a single package can be noticeable O(N*N).

I don't have a solution that I like right now, unfortunately.

cristaloleg: We can invert loop below, so instead of creating a map and iterating over params, we can iterate over params first and check if it was used.

Quasilyte: If you can make it work, send an updated patch set.
`types.Info.Uses` can hold huge amount of entries for some packages, I remember this being an issue in some of my previous projects, so we definitely want to avoid looping through it more than once.

Quasilyte:
Also note that you may choose to merge this "as is", if "experimental" will be set to true. :)

cristaloleg:
Sounds good for me

lint/unusedParam_checker.go
```diff
@@ -0,0 +1,60 @@
...
+	if params == nil || params.NumFields() == 0 {
+		return
+	}
+
+	// collect all params to map
+	objToIdent := make(map[*ast.Object]*ast.Ident)
+	for _, p := range params.List {
+		if len(p.Names) == 0 {
+			c.warnUnnamed(p)
+			break
```
Quasilyte: I believe it can be more sensible to do a `return` here?

lint/unusedParam_checker.go
```diff
@@ -0,0 +1,60 @@
...
+		}
+		for _, id := range p.Names {
+			if id.Name != "_" {
+				objToIdent[id.Obj] = id
+			}
+		}
+	}
+
+	// remove used params
+	for id := range c.ctx.TypesInfo.Uses {
```
Quasilyte: We still do full `Uses` range here.
It can be more efficient to walk function body and mark params as used along the way,
looking up only usages present inside function itself.
Right now it ranges over whole package Uses set for every function.

If you don't feel like solving this issue right now, we can add this checker as experimental
along with TODO that explains that there is potential performance issue there.

lint/unusedParam_checker.go
cristaloleg: Yep, can be. Good catch.

lint/unusedParam_checker.go
cristaloleg: Well, below 1M(even 10M) iterations we're safe. I don't think this will explode in near future.
Sure, we can make an issue for that, but...I doubt that someone will mention a problem regarding this loop.
😉

lint/unusedParam_checker.go
```diff
@@ -0,0 +1,61 @@
+package lint
+
+import (
+	"go/ast"
+)
+
+func init() {
+	addChecker(unusedParamChecker{}, &ruleInfo{})
```
Quasilyte: Add `Experimental: true` here, and we can merge it.

PR #236, opened 2018-06-16
[WIP] lint: implementation assert in code files

### LINKED ISSUE #206: lint: add check for interface implementation in code files
before:
```
// code.go
var _ pkg.MyInterface = (*MyStruct)(nil)
```
after:
```
// code_test.go
var _ pkg.MyInterface = (*MyStruct)(nil)
```
This will prevent unnecessary dependency on `pkg` package in code files, rather than tests.

cristaloleg:
Fixes #206


lint/implAssert_checker.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+		if _, ok := c.ctx.TypesInfo.TypeOf(valueSpec.Type).(*types.Interface); !ok {
+			return
+		}
+		if len(valueSpec.Values) != 1 {
+			return
+		}
+		if _, ok := c.ctx.TypesInfo.TypeOf(valueSpec.Values[0]).(*types.Struct); !ok {
+			return
+		}
```
Quasilyte: For this kind of early return sequences, we sometimes use this idiom:
```diff
- 	if !ok {
- 		return
- 	}
- 	if len(valueSpec.Names) != 1 || valueSpec.Names[0].Name != "_" {
- 		return
- 	}
- 	if _, ok := c.ctx.TypesInfo.TypeOf(valueSpec.Type).(*types.Interface); !ok {
- 		return
- 	}
- 	if len(valueSpec.Values) != 1 {
- 		return
- 	}
- 	if _, ok := c.ctx.TypesInfo.TypeOf(valueSpec.Values[0]).(*types.Struct); !ok {
- 		return
- 	}
+ cond := ok &&
+ 	len(valueSpec.Names) == 1 &&
+ 	len(valueSpec.Values) == 1 &&
+ 	valueSpec.Names[0].Name == "_"
+ if !cond {
+ 	return
+ }
+ if _, ok := c.ctx.TypesInfo.TypeOf(valueSpec.Type).(*types.Interface); !ok {
+ 	return
+ }
+ if _, ok := c.ctx.TypesInfo.TypeOf(valueSpec.Values[0]).(*types.Struct); !ok {
+ 	return
+ }
```

If we had `typep`:
```go
cond := ok &&
	len(valueSpec.Names) == 1 &&
	len(valueSpec.Values) == 1 &&
	valueSpec.Names[0].Name == "_" &&
	typep.IsInterface(c.ctx.TypesInfo.TypeOf(valueSpec.Type)) &&
	typep.IsStruct(c.ctx.TypesInfo.TypeOf(valueSpec.Values[0]))
if !cond {
	return
}
```

Quasilyte:
Maybe it's OK to call positive_tests.go package without `_test` suffix, so you can catch warnings there?

cristaloleg:
It's not prepared for a review yet 😉

PR #240, opened 2018-06-17
all: VeryOpinionated checker's flag

cristaloleg:


lint/lint.go
```diff
@@ -15,19 +15,23 @@ type checkFunction interface {
...
 	//	- rule requires more testing or alternative design
 	Experimental bool
 
-	// SyntaxOnly marks rules that can be checker by using AST only.
-	// Such rule implementations should not use types info.
-	//
-	// It's OK not to mark rule as SyntaxOnly even if current
-	// implementation does not use types info.
-	SyntaxOnly bool
+	// VeryOpinionated marks rule as controversive for some audience and
```
Quasilyte: s/controversive/controversial/g (there is another case below)?

cristaloleg: indeed, my bad

PR #241, opened 2018-06-17
lint: add boolFuncPrefix checker

### LINKED ISSUE #230: lint/boolFuncPrefix: consider to name bool functions with `Is` prefix
name: `boolFuncPrefix `
before:
```go
func Enabled() bool
```
after:
```go
func IsEnabled() bool
```

cristaloleg:
Fixes #230


Quasilyte:
What about `boolFuncPrefix` (`boolFuncPrefixChecker`)?
Flag makes it somewhat confusing because of `flag` package.

cristaloleg:
Agreed. I'm not so good at naming things sometimes :(

Quasilyte:
I also suggesting trying this out on GOROOT/src and catching corner cases that may don't need suggested prefix.

It will have many false positives for functions that are not predicates, but return bool to specify whether operation failed or not. Like `doRequest` or something like that. `isDoRequest` is not the right thing there (or is it?).

lint/flagFuncNaming_cheker.go
```diff
@@ -0,0 +1,56 @@
...
+func (c *flagFuncNamingChecker) warn(fn *ast.FuncDecl) {
+	c.ctx.Warn(fn, "consider to add Is/Has/Contains prefix to function name")
+}
+
+func (c *flagFuncNamingChecker) isBoolType(expr ast.Expr) bool {
+	switch typ := c.ctx.TypesInfo.TypeOf(expr).(type) {
+	case *types.Basic:
+		return typ.Kind() == types.Bool
+	default:
+		return false
```
Quasilyte: (optional.)
Can do an `if` for a single case:
```go
if typ, ok := c.ctx.TypesInfo.TypeOf(expr).(*types.Basic); ok {
	return typ.Kind() == types.Bool
}
return false
```

Quasilyte: Even simpler:
```go
typ, ok := c.ctx.TypesInfo.TypeOf(expr).(*types.Basic)
return ok && return typ.Kind() == types.Bool
```

lint/boolFuncPrefix_cheker.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"go/ast"
+	"go/types"
+	"strings"
+)
+
+func init() {
+	addChecker(boolFuncPrefixChecker{}, &ruleInfo{
+		Experimental: true,
+	})
```
Quasilyte: `addChecker(boolFuncPrefixChecker{}, attrExperimental, attrVeryOpinionated)`

Quasilyte:
Thanks for doing some investigation on other projects (I see new prefixes there).
LGTM when build is fixed.

PR #242, opened 2018-06-17
lint: re-design attribute passing to addChecker function

Quasilyte:
This makes passing empty ruleInfo not needed.
We retain type-safety and all benefits of "false" by
default of original approach.

Also mark unexportedCall as VeryOpinionated while there.

### DIFF
--- a/lint/attributes.go
+++ b/lint/attributes.go
@@ -5,9 +5,8 @@ package lint
 
 type checkerAttribute int
 
-// All valid attributes.
-var (
-	attrExperimental    = new(checkerAttribute)
-	attrSyntaxOnly      = new(checkerAttribute)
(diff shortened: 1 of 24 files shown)


lint/attributes.go
```diff
@@ -0,0 +1,13 @@
+package lint
+
+// TODO(quasilyte): move attribute-related things here
+// when this refactoring is complete.
+
+type checkerAttribute int
+
+// All valid attributes.
+var (
+	attrExperimental    = new(checkerAttribute)
```
cristaloleg: To be honest I will expect flags, not a pointer. So:
```
attrExperimental checkerAttribute =  1 << iota
attrSyntaxOnly
attrVeryOpinionated
```
But it's kinda a minor change. so np

PR #250, opened 2018-06-17
all: consolidate linter registration

cristaloleg:


Quasilyte:
I think we should think hard about this part with `New` method that both accepts checker and returns wrapped checker (new instance).
Maybe we can, at least, use receiver instance and call method like `Bind`?
Example:
```diff
-func (c *boolFuncPrefixChecker) New(ctx *context) func(*ast.File) {
- 	return wrapFuncDeclChecker(&boolFuncPrefixChecker{
- 		baseFuncDeclChecker: baseFuncDeclChecker{ctx: ctx},
- 	})
-}
+func (c *boolFuncPrefixChecker) New(ctx *context) func(*ast.File) {
+	c.baseFuncDeclChecker = baseFuncDeclChecker{ctx: ctx}
+	return wrapFuncDeclChecker(c)
+}
```
I mean, we already have instance, why not use it instead of creating another instance?

Quasilyte:
Other method name options: `CheckFunction` (because it returns check function) or `Wrap` (because it usually does wrapping).

PR #252, opened 2018-06-17
docs: get docs from checker comments

fexolm:


fexolm:
Ready for appendCombine, please check

cmd/makedocs/main.go
```diff
@@ -2,48 +2,73 @@ package main
...
+			continue
+		}
+		c := checker{
+			Name:            r.Name(),
+			SyntaxOnly:      r.SyntaxOnly,
+			Experimental:    r.Experimental,
+			VeryOpinionated: r.VeryOpinionated,
+		}
+		for _, comment := range f.Comments {
+			if strings.HasPrefix(comment.Text(), "!") {
```
Quasilyte: `Text()` computes combined text every time you call it.
Maybe it's enough to check first comment from group test for that prefix?
Or re-use `Text()` result.
Just making sure you know what's going on here. There should be no real performance issue here.

cmd/makedocs/main.go
```diff
@@ -59,15 +84,62 @@ func main() {
...
-	b, err := ioutil.ReadFile(checkersPath + name + ".md")
-	return string(b), err
-}
+type parseState uint
+
+const (
+	emptyLine parseState = 1 << iota
+	desciption
+	before
+	after
```
Quasilyte: Usually, we prefix enum values with some common prefix that resembles their type.
`stEmptyLine`, `stDescriotion`, etc.
Yeah, like in C. :)

cmd/makedocs/main.go
```diff
@@ -59,15 +84,62 @@ func main() {
...
+)
+
+func parseComment(text string, c *checker) {
+	lines := strings.Split(text, "\n")
+	s := emptyLine
```
Quasilyte: I would prefer `s/s/st/` to avoid confusion with "string".
(Optional.)

cmd/makedocs/main.go
cristaloleg: From the other side: how many comments it will process? Hundreds at most ;)

Quasilyte: Yep, this is why it doesn't matter.
Just making a note that `Text()` is not a simple getter.

Quasilyte:
What about more explicit state machine?
(pseudo-go):
```go
grammar := []syntax{
  stx.Summart,
  stx.Or(stx.All(stx.EmptyLine, stx.Details, stx.EmptyLine), stx.EmptyLine),
  stx.Before,
  stx.EmptyLine,
  stx.After,
  stx.Or(stx.All(stx.EmptyLine, stx.Notice, stx.End), stx.End),
}
```
Each phase captures state and can fill doc being generated.
It's also possible to print which exact phase failed, during which alternation, which element was expected, etc.

(This is only an idea, not request for re-implementation.)

fexolm:
@Quasilyte @cristaloleg please watch

cmd/makedocs/main.go
```diff
@@ -59,15 +86,82 @@ func main() {
...
+	return nil
 }
 
-func shortDescription(desc string) string {
-	// We expect every checker to have first doc line
-	// to be a short summary, complete sentence.
-	summary := strings.Split(desc, "\n")[0]
-	summary = strings.TrimSpace(summary)
-	return summary
+func parseDesc(lines []string, ind *int, c *checker) error {
```
cristaloleg: s/ind/index/g 😉

cmd/makedocs/main.go
```diff
@@ -12,38 +17,60 @@ import (
...
+		if !ok {
+
+			log.Printf("File not found: %s%s_checker.go", checkersPath, r.Name())
+			continue
+		}
+		c := checker{
+			Name:            r.Name(),
+			SyntaxOnly:      r.SyntaxOnly,
+			Experimental:    r.Experimental,
+			VeryOpinionated: r.VeryOpinionated,
```
Quasilyte: You could store `*lint.Rule` inside `checker` and do just: `c := checker{rule: rule}`.
Current approach also makes sense, so this is not a request for changes.

cmd/makedocs/main.go
```diff
@@ -59,15 +86,82 @@ func main() {
...
+		return errors.New("parseDesc: no description provided")
+	}
+	if strings.HasPrefix(lines[*index], "@") { // if no description
+		return nil
+	}
+	for *index < len(lines) && len(lines[*index]) > 0 {
+		c.Description += strings.TrimSpace(lines[*index]) + "\n"
+		*index++
+	}
+	*index++ //skip empty line
```
Quasilyte: You could do every stage parsing function a method of some type, which would store index inside itself,
and do something like:
```go
p.skipEmptyLine()
```
Which would be more readable and can also perform a check that current like is really empty or not.

Quasilyte: Can be refactored later.

cmd/makedocs/main.go
```diff
@@ -59,15 +86,82 @@ func main() {
...
+func parseBefore(lines []string, index *int, c *checker) error {
+	if len(lines) <= *index || strings.TrimSpace(lines[*index]) != "@Before:" {
+		return errors.New("parseBefore: no @Before: section found")
+	}
+	*index++
+	for *index < len(lines) && len(lines[*index]) > 0 {
+		c.Before += strings.TrimSpace(lines[*index]) + "\n"
+		*index++
+	}
+	*index++ //skip empty line
```
Quasilyte: Nit: no space between `//` and comment text.
```diff
- //skip empty line
+ // skip empty line
```
I can fix that for you.

Quasilyte:
Also, thank you for doing this. :)

PR #253, opened 2018-06-17
Quasilyte/gosu pro lint redesign

Quasilyte:
Main changes:
- 1 checker base `checkerBase` instead of N checker bases. There are still N interfaces and we need them both for documentation and type assertions inside `addChecker`, but having multiple bases was a mistake. It lead to some code duplication and higher overall complexity.
- No need to bind context inside checker code. Checkers that only did context assignment do not need `New` function anymore. Checkers that need initalization can define `Init()` function for that.
- `wrapXxx` call is done by `addChecker`, so checkers are not required to wrap themselves anymore.

My main objective was to simplify checkers code even if this required some complications inside `addChecker` function. Also, it's simpler now to choose base type -- there is only one correct option.
It is also simpler to add new walkers/wrappers/whether we call them, since now there is no need to add new base type that implements `VisitFunc`/`PerFuncInit`, etc.

We also can modify our implementation of this magic later as API should be enough to encapsulate that completely.

### DIFF
--- a/lint/checker_base.go
+++ b/lint/checker_base.go
@@ -6,28 +6,34 @@ import (
 )
 
 type (
+	// walkerEventHandler describes common hooks available for every checker.
+	walkerEventHandler interface {
+		// VisitFunc is called for every function declaration that is about
+		// to be traversed. If false is returned, function is not visited.
+		VisitFunc(*ast.FuncDecl) bool
+	}
+
 	// funcDeclChecker visits every top-level function declaration.
 	//
-	// See also: baseFuncDeclChecker, wrapFuncDeclChecker.
+	// See also: wrapFuncDeclChecker.
 	funcDeclChecker interface {
+		walkerEventHandler
 		CheckFuncDecl(*ast.FuncDecl)
 	}
 
 	// exprChecker visits every expression inside AST file.
 	//
-	// See also: baseExprChecker, wrapExprChecker.
+	// See also: wrapExprChecker.
 	exprChecker interface {
+		walkerEventHandler
 		CheckExpr(ast.Expr)
 	}
 
 	// localExprChecker visits every expression inside function body.
 	//
-	// PerFuncInit is called for every function visited.
-	// If returned false, function is skipped.
-	//
-	// See also: baseLocalExprChecker, wrapLocalExprChecker.
+	// See also: wrapLocalExprChecker.
 	localExprChecker interface {
-		PerFuncInit(*ast.FuncDecl) bool
+		walkerEventHandler
 		CheckLocalExpr(ast.Expr)
 	}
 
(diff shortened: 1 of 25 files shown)


Quasilyte:
I'm pretty happy with this design.
Implementation is somewhat complicated, but it can be changed later if anyone has a better idea **without** making checker creation guide invalidated.

Quasilyte:
`cloneAbstractChecker` is needed to perform create new zero value of passed value.
This is required to make `BindContext`+`Init` work on a fresh copy of underlying checker.
With this, `New` will return independent checker => makes it possible to have multiple checkers for the same rule => easier to implement thread-local checkers => can run same check in parallel.
See this example https://play.golang.org/p/YO5p3C1_QmI
If we were assigning `b := a` without `newZeroValue`, both instances would be modified.

lint/checker_base.go
```diff
@@ -57,25 +61,37 @@ type (
...
-type baseFuncDeclChecker struct {
+type checkerBase struct {
 	ctx *context
 }
 
+func (c *checkerBase) BindContext(ctx *context) {
+	c.ctx = ctx
+}
+
+func (c *checkerBase) Init() {}
```
cristaloleg: What about a comment that states that `Init` should be implemented in checker to initialise some data?

lint/lint_private.go
```diff
@@ -8,20 +8,31 @@ import (
...
+// embedding checker into addChecker.
+//
+// It's an implementation detail that has only these guarantees:
+//	- implemented by checkerBase type
+//	- it is an argument type for addChecker
+//
+// Therefore, every checker that embeds checkerBase (normally, every checker)
+// also a valid argument to addChecker.
+type abstractChecker interface {
+	BindContext(*context)
```
cristaloleg: Nice name

lint/lint_private.go
```diff
@@ -51,22 +62,65 @@ func (ctx *context) Warn(node ast.Node, format string, args ...interface{}) {
...
-	ruleName := typeName[len("*lint.") : len(typeName)-len("Checker")]
-	info.New = c.New
-	checkFunctions[ruleName] = &info
+
+	proto := checkerProto{rule: &rule}
+	proto.clone = func(ctx context) *Checker {
+		c := cloneAbstractChecker(c)
+		clone := Checker{Rule: proto.rule}
+		clone.check = bindCheckFunction(c)
+		clone.ctx = ctx
```
cristaloleg: WDYT?
```
clone := Checker{
Rule: proto.rule,
check: bindCheckFunction(c),
ctx: ctx,
}
```

Quasilyte: SGTM.

lint/checker_base.go
Quasilyte: Also commented on BindContext and added a "See checkerBase" for abstractChecker type comment.

lint/lint_private.go
```diff
@@ -8,20 +8,29 @@ import (
...
+	// clone performs prototype copy and returns it as *Checker.
+	clone func(context) *Checker
 }
 
-type checkFunction interface {
-	New(*context) func(*ast.File)
+// abstractChecker is a proxy interface to forward checkerBase
+// embedding checker into addChecker.
+//
+// abstarctChecher is implemented by checkerBase directly and completely,
```
cristaloleg: s/abstarctChecher/abstractChecker :P

Quasilyte: Good catch.

PR #254, opened 2018-06-18
cmd/gocritic: add checker-new command

Quasilyte:


Quasilyte:
I'll fix lint error later. This is not high priority and can be delayed.

cmd/gocritic/main.go
```diff
@@ -34,6 +35,11 @@ var subCommands = []*subCommand{
 		name:  "check-project",
 		short: "run gocritic over specified source tree, recursively",
 	},
+	{
+		main:  checkertools.NewMain,
+		name:  "checker-new",
```
cristaloleg: Not PR related, but: what about renaming commands to `package`, `project`, `newlinter` ?
Till we've some time before public release.

cmd/checkertools/new.go
```diff
@@ -0,0 +1,126 @@
...
+		log.Printf("usage: rulename")
+		flag.PrintDefaults()
+	}
+	flag.Parse()
+
+	if flag.NArg() != 1 {
+		flag.Usage()
+		os.Exit(1)
+	}
+	// TODO: add camel case check.
```
cristaloleg: I think we can add note in usage about camel case style and if name doesn't fit proper style we can cover this during review before merging 😉

cmd/checkertools/new.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	return testDir
+}
+
+func createPositiveTests(testDir string) {
+	const data = `package checker_test
+
+// TODO: write tests that trigger newly added checker.
+
+/// this is example of warning directive. It fails your tests right now
```
cristaloleg: What about note about `///` ? 😉

cristaloleg: Or explain what this line should contain, ex warning message.

cmd/checkertools/new.go
```diff
@@ -0,0 +1,126 @@
...
+// Optional section for details.
+//
+// Before:
+// badCode := "minimal example of code that triggers a warning"
+//
+// After:
+// goodCode := "example of intended fix"
+
+func init() {
+	addChecker(&{{.TypeName}})
```
cristaloleg: What about comments regarding attributes? and notice to remove it 😉

cristaloleg: However it's not important.

cmd/gocritic/main.go
Quasilyte: I'm not sure. Can't alias solve problems for users that don't like typing extra `len("check-")`?

cmd/checkertools/new.go
Quasilyte: We should decide whether this is a tutorial or stub generator.
Or add `-with-help` flag that is true by default.

dmirogin:
Does we really want to make this command hmm..., public?
When somebody install gocritic then he will see checker-new command together with other helpfull for user commands ?

Quasilyte:
@dmirogin, agreed.
Also, it's probably simpler to do it from scratch since there were multiple big changes since this PR.

PR #257, opened 2018-06-18
lint: fix unnamed recv issue with localNameChecker

Quasilyte:
Also add sanity test to detect such problems earlier in future.


Quasilyte:
It's like a minefield.

PR #266, opened 2018-06-18
lint:unusedParam: make early exit

### LINKED ISSUE #263: lint: unusedParam is very slow
This is not a bug as `unusedParam` is marked as experimental, but still, this the performance issue is real.

Run:
```
time gocritic check-project -enable=unusedParam $GOROOT/src
```
Wait for it... wait for it.. Have a cup of tea.
```
real	3m15.086s
user	3m37.088s
sys	0m3.508s
```

This is because there is full `types.Info.Uses` traversal per every function declaration.

cristaloleg:
Fixes #263


Quasilyte:
Are you sure it fixes it?
It still can do whole Uses traversal in the worst case as map iteration order is somewhat randomized.

Issue #261, opened 2018-06-18
lint: docStub false positives


Quasilyte:
Reproducer:
```
gocritic check-project -enable=docStub $GOROOT/src
```
Output:
```
cmd/compile/internal/syntax/parser.go:2026:1: docStub: silencing go lint doc-comment warnings is unadvised
```
Code:
```go
// Statement =
// 	Declaration | LabeledStmt | SimpleStmt |
// 	GoStmt | ReturnStmt | BreakStmt | ContinueStmt | GotoStmt |
// 	FallthroughStmt | Block | IfStmt | SwitchStmt | SelectStmt | ForStmt |
// 	DeferStmt .
func (p *parser) stmtOrNil() Stmt {
```
This is not doc comment stub, and it's not exported function at all.
To avoid such false positives, `docStub` should check that function being checked is actually exported (see `ast.IsExported`) and check that there are no more lines inside godoc comment.
Comment above has 5 elements in it's comment group.

Negative tests must be expanded.

Quasilyte:
CC @weeellz, are you interested in this?

cristaloleg:
I can pick up this one :)

PR #268, opened 2018-06-19
lint/docStub: ignore unexported functions

cristaloleg:
Fixes and updates #261

### DIFF
--- a/cmd/gocritic/testdata/docStub/positive_tests.go
+++ b/cmd/gocritic/testdata/docStub/positive_tests.go
@@ -1,7 +1,19 @@
 package checker_test
 
+type kek struct{}
+
 // Foo ...
 /// silencing go lint doc-comment warnings is unadvised
-func Foo() {
+func Foo() {}
 
-}
+// Wow
+/// silencing go lint doc-comment warnings is unadvised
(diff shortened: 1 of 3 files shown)


cmd/gocritic/testdata/docStub/positive_tests.go
```diff
@@ -1,7 +1,19 @@
...
+func (k *kek) Wow() {}
 
-}
+//Well
+/// silencing go lint doc-comment warnings is unadvised
+func Well() {}
+
+// Goo go go.
+/// silencing go lint doc-comment warnings is unadvised
+func Go() {}
```
Quasilyte: This doesn't look like something `docStub` should report as "silencing go lint".
It's also not a valid doc comment and go lint will trigger on it.
`docStub` aims to find comments that are obviously "hacks" to avoid go lint complaints.
Like:
```go
// Foo ...
func Foo() {}
```
`go lint` would not trigger on this, but `docStub` should. It's a tool that makes it easier to refactor such code that used doc comment stubs.

Quasilyte:
Ping.

If you want to turn `docStub` into something different, perhaps it's probably a good idea to describe intended functionality in a new issue?

PR #282, opened 2018-06-21
cmd/lintwalk/main: prevent panic in filepath.Walk

olshevskiy87:


Quasilyte:
Issue reproduced, thank you.
It does panic when there is no such folder, for example.
With this patch, it does terminate and print error message as it should:
```
check-project: walk src-root: lstat path/to/something/blah: no such file or directory
```

Issue #280, opened 2018-06-21
cmd/makedocs: misformatted code blocks


Quasilyte:
Given this pattern:
```go
//! Detects when predeclared identifiers shadowed in assignments.
//
// @Before:
// func main() {
//     // shadowing len function
//     len := 10
//     println(len)
// }
//
// @After:
// func main() {
//     // change identificator name
//     length := 10
//     println(length)
// }
```
This is expected output for "Before" section:
![image](https://user-images.githubusercontent.com/6286655/41727334-c606e35a-757c-11e8-907f-ef27215402a0.png)
This is what we get:
![image](https://user-images.githubusercontent.com/6286655/41727364-d8255e5e-757c-11e8-8206-09613e95aaa0.png)

"After" section is has the same issue as well.

PR #283, opened 2018-06-21
cmd/makedocs: fix trimmed spaces in code samples

fexolm:


Quasilyte:
Hmm.. Maybe we should trim spaces from summary, note and details, but not from `@before` and `@after`?
Ideally, we would signal an error if any of non-code section does not meet our "perfect doc" criteria.
For now, this PR does solve most critical part and fixes #280.

fexolm:
@Quasilyte, i think it doesn't really matter if we would trim only `@before` and `@after` or all sections. Maybe there is some cases where users will need custom indents in `@note` for example.

PR #285, opened 2018-06-21
docs/overview.md.tmpl: summary section is missing from the output

### LINKED ISSUE #281: cmd/makedocs: summary section is missing from the output
Given this pattern:
```
//! Detects when predeclared identifiers shadowed in assignments.
//
// @Before:
// func main() {
//     // shadowing len function
//     len := 10
//     println(len)
// }
//
// @After:
// func main() {
//     // change identificator name
//     length := 10
//     println(length)
// }
```
We get markdown document which does not include `Detects when predeclared identifiers shadowed in assignments` string at all.

Here is diff with older docs:
```diff
 ## builtinShadow
-Detects when predeclared identifiers shadowed in assignments.
+
```

fexolm:
fixes #281

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -162,8 +162,6 @@ Detects `append` chains to the same slice that can be done in a single `append` 
 
 
 
-
-
 **Before:**
 ```go
 xs = append(xs, 1)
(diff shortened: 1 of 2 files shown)


Quasilyte:
Commit message prefix should be like `cmd/docs: ...`, because it affects multiple files.
At least I think it's clearer.

Quasilyte:
Could you add `Fixes #281` into commit message body?

Issue #231, opened 2018-06-16
lint: ensure SyntaxOnly checkers can't use types.Info


quasilyte:
When checker marked as SyntaxOnly tries to access types.Info from context,
it should panic to make this contract violation apparent.

See also https://github.com/go-critic/go-critic/pull/229.

cristaloleg:
Do you've any ideas how to implement this?

Quasilyte:
Yes, at least 2. But I don't like either of them that much.

PR #286, opened 2018-06-21
add imported package names shadowing linter

NexoMichael:
Sometimes developers use the same name for the variables as imported packages, more often I see something like importing "log" package and having "log" variable in the code.

This linter checks what imports the package has and what variables got created and compares one to another.

### DIFF
--- a/lint/importShadow_checker.go
+++ b/lint/importShadow_checker.go
@@ -34,7 +34,7 @@ import (
 )
 
 func init() {
-	addChecker(&importShadowChecker{}, attrSyntaxOnly)
+	addChecker(&importShadowChecker{}, attrExperimental, attrSyntaxOnly)
 }
 
 type importShadowChecker struct {
--- a/lint/importShadow_checker.go
+++ b/lint/importShadow_checker.go
@@ -1,0 +1,61 @@
+package lint
+
+//! Detects when imported package names shadowed in assignments.
+//
+// @Before:
+// import (
+// 	   "fmt"
+//     "math"
+// )
+// func main() {
+//     fmt.Println(math.Pi)
+//     // shadowing math package
+//     math := 10
+//     fmt.Println(len)
+// }
+//
+// @After:
+// import (
+//     "fmt"
+//     "math"
+// )
+// func main() {
+//     fmt.Println(math.Pi)
+//     // change identificator name
+//     value := 10
+//     fmt.Println(value)
+// }
+
+import (
+	"go/ast"
+	"go/types"
+
+	"github.com/go-critic/go-critic/lint/internal/astwalk"
+)
+
+func init() {
+	addChecker(&importShadowChecker{}, attrSyntaxOnly)
+}
+
+type importShadowChecker struct {
+	checkerBase
+}
+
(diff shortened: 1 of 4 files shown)


Quasilyte:
My favorite case is `path/filepath` and `path`.
They make me nervous when naming something that holds file path. :sweat_smile:

lint/importShadow_checker.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"go/ast"
+	"go/types"
+
+	"github.com/go-critic/go-critic/lint/internal/astwalk"
+)
+
+func init() {
+	addChecker(&importShadowChecker{}, attrSyntaxOnly)
```
Quasilyte: Since this checker uses `types.Package`, it should probably not have `attrSyntaxOnly` attribute.
Generally, it's safer to omit this attribute, at least until we implement #231.

lint/importShadow_checker.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+type importShadowChecker struct {
+	checkerBase
+}
+
+func (c *importShadowChecker) Init() {}
```
Quasilyte: You may remove `Init` method as this checker has no state that needs to be initialized.

fexolm:
Please add `attrExperimental` to you checker. We always mark new checkers as experemental.

lint/importShadow_checker.go
```diff
@@ -0,0 +1,61 @@
+package lint
+
+//! Detects when imported package names shadowed in assignments.
+//
+// @Before:
+// import (
+// 	   "fmt"
```
Quasilyte: Looks like there is an extra whitespace here.

NexoMichael: yeap, but extra white space will need some fix in docs generation, will look what is going on there

NexoMichael: let's create additional issue for that, it seems like current approach with
```
for *index < len(lines) && len(lines[*index]) > 0 {
```
for finding description parts will not work for code with empty lines

Quasilyte:
Note: we're following golang/go guidelines for commit messages.

tl;dr: it should be like: `lint: add importShadow linter`
And the body usually contains either details or `Updates #XXX`/`Fixes #XXX` to provide more context.

lint/importShadow_checker.go
fexolm: Its how our docs works. Empty lines in code are not allowed.

Quasilyte:
@NexoMichael, you can use your checker either with `-withExperimental` option that turns on all experimental checkers for `-enable=all` or with explicit list like `-enable=importShadow`.

lint/importShadow_checker.go
Quasilyte: This is mostly because we want:
1. Encourage minimal examples.
2. Avoid hard to read doc comments inside checkers source files.

I would use example like this:
```go
// "math" and "fmt" packages are imported.
func f(math int) int {
  fmt := math+1
  return fmt
}
```
Or even something that does not include surrounding func.

PR #287, opened 2018-06-21
docs: fix importShadow docs

fexolm:
CC @NexoMichael

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -680,7 +680,7 @@ Detects when imported package names shadowed in assignments.
 **Before:**
 ```go
 import (
-	   "fmt"
+    "fmt"
     "math"
 )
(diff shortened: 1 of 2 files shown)


Quasilyte:
CC @NexoMichael (just for notification)

Quasilyte:
Have we decided whether code snippets should use spaces or tabs?
Tabs are closer to real Go code, but spaces are somewhat simpler to type (in most editors/IDEs).
We can open an issue for this.

Quasilyte:
Ideally, code snippets that are not properly formatted should result in makedocs error.
This way, there would be no such issue from the first place.
`go/format` could help here to apply gofmt-like formatting on before/after sections and check output with input code.

PR #289, opened 2018-06-21
all: ability to skip auto generated files

### LINKED ISSUE #247: all: ability to not scan autogenerated files
We should easily ignore generated go files(with `DO NOT EDIT` label at the top), in particular we might be also interested in ignoring `protobuf` files.

cristaloleg:
Fixes #247

### DIFF
--- a/cmd/criticize/main.go
+++ b/cmd/criticize/main.go
@@ -185,10 +185,7 @@ func (l *linter) CheckPackage(pkgPath string) {
 }
 
 func isGenerated(f *ast.File) bool {
-	if f.Comments == nil || len(f.Comments) == 0 {
-		return false
-	}
-	return generatedFileCommentRE.MatchString(f.Comments[0].Text())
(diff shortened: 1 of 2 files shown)


cmd/criticize/main.go
```diff
@@ -169,16 +177,25 @@ func (l *linter) CheckPackage(pkgPath string) {
...
-		l.checkFile(f)
+		if l.checkGenerated || !isGenerated(f) {
+			l.ctx.SetFileInfo(l.getFilename(f))
+			l.checkFile(f)
+		}
+	}
+}
+
+func isGenerated(f *ast.File) bool {
+	if f.Comments == nil || len(f.Comments) == 0 {
```
Quasilyte: Isn't nil check redundant here?

cristaloleg: Indeed. Hm...sounds like an idea for check. Will fill an issue ;D

Quasilyte: gosimple should catch this. I wonder why it doesn't trigger on this one.
Almost feeling uncomfortable because of this.

PR #293, opened 2018-06-21
lint/lenNil: detect redundant x!=nil before len(x)

### LINKED ISSUE #291: lint/lenNil: detect redundant x!=nil before len(x)
Before:
```go
if myMap != nil && len(myMap) == smth { 
```
After:
```go
if len(myMap) == smth { 
```
See: https://golang.org/pkg/builtin/#len

cristaloleg:
Fixes #291 

TODO:
- Support not only if's but also switch, returns so any stmt. - done
- Consider `x == nil` as redundant too.  - done

### DIFF
--- a/cmd/gocritic/testdata/lenNil/negative_tests.go
+++ b/cmd/gocritic/testdata/lenNil/negative_tests.go
@@ -1,11 +1,6 @@
 package checker_test
 
-func g1() bool {
-	var ch chan int
-	return ch != nil && len(ch)+10 == 100
-}
-
-func g2() {
+func g1() {
 	var ch chan int
 	if ch == nil {
 		return
@@ -13,22 +8,17 @@ func g2() {
 	if len(ch) == 0 {
 		return
 	}
-
-	switch ch != nil && len(ch)+10 == 100 {
-	default:
-	}
-
-	switch {
-	case ch != nil || len(ch) == 0:
-	case ch != nil && len(ch) == 0:
-	default:
-	}
 }
 
-func BUG() {
-	var ch chan int
-	/// ch == nil check is redundant
-	/// ch != nil check is redundant
-	if (ch == nil || len(ch) == 1) != (ch != nil && len(ch) != 1) {
+func g2() {
+	var x map[int]int
+	var y []int
+	if x != nil && len(y) != 0 {
+	}
+	if x != nil && len(y) != 0 {
+	}
+	if len(y) != 0 || x != nil {
+	}
+	if 0 != len(y) || nil != x {
 	}
 }
--- a/cmd/gocritic/testdata/lenNil/negative_tests.go
+++ b/cmd/gocritic/testdata/lenNil/negative_tests.go
@@ -1,0 +1,25 @@
+package checker_test
+
+func g1() bool {
+	var m map[int]int
+	if m == nil && len(m) == 0 {
+		return false
+	}
+	var ch chan int
+	return ch != nil && len(ch)+10 == 100
+}
+
+func g2() {
+	var ch chan int
+	if ch == nil {
+		return
+	}
+	if len(ch) == 0 {
+		return
+	}
+
+	switch ch != nil && len(ch)+10 == 100 {
+	case true:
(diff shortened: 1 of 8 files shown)


cmd/gocritic/testdata/lenNil/positive_tests.go
```diff
@@ -0,0 +1,24 @@
+package checker_test
+
+func f() {
+	{
+		var m map[int]int
+		/// consider to remove redundant nil check
```
Quasilyte:
```diff
- consider to remove redundant nil check
+ m != nil check is redundant
```
?
We can easily print AST nodes with `%s` inside formatting string.

Quasilyte: Can be improved in a separate PR as an improvement.

lint/lenNil_checker.go
```diff
@@ -0,0 +1,71 @@
...
+// if len(myMap) == smth {
+
+import (
+	"go/ast"
+	"go/token"
+	"go/types"
+)
+
+func init() {
+	addChecker(&lenNilChecker{})
```
Quasilyte: Should be marked as experimental.
Please, add `attrExperimental` to the `addChecker` call.

lint/lenNil_checker.go
```diff
@@ -0,0 +1,71 @@
+package lint
+
+//! Detects redundant x!=nil before len(x).
+//
+// @Before:
+// if myMap != nil && len(myMap) == smth {
+//
+// @After:
+// if len(myMap) == smth {
```
Quasilyte: To avoid bad syntax highlighting, maybe we can change `{` to `{}`?

cmd/gocritic/testdata/lenNil/negative_tests.go
```diff
@@ -0,0 +1,25 @@
...
+		return
+	}
+	if len(ch) == 0 {
+		return
+	}
+
+	switch ch != nil && len(ch)+10 == 100 {
+	case true:
+	default:
+	}
```
Quasilyte: What about `x != nil || len(x) == 0`?
I don't see `Tok == token.LAND` check inside the implementation.

lint/lenNil_checker.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	expr, ok := ifstmt.Cond.(*ast.BinaryExpr)
+	if !ok {
+		return
+	}
+
+	if (c.isNilCheck(expr.X) && c.hasLenCall(expr.Y)) ||
+		(c.isNilCheck(expr.Y) && c.hasLenCall(expr.X)) {
+		c.warn(ifstmt)
```
Quasilyte: Should probably warn on binary expr that contains redundant nil check.

cmd/gocritic/testdata/lenNil/negative_tests.go
Quasilyte: Also, I don't see checks for `len` argument.
This code may result in a false positive: `x != nil && len(y) == 0`.

lint/lenNil_checker.go
```diff
@@ -60,12 +74,17 @@ func (c *lenNilChecker) isRefType(expr ast.Expr) bool {
 	}
 }
 
+func (c *lenNilChecker) isNilKeyword(x ast.Expr) bool {
+	id, ok := x.(*ast.Ident)
+	return ok && id.Name == "nil"
```
Quasilyte: Could simplify to `qualifiedName(x) == "nil"`.
Also, `nil` is not a keyword. :)

Quasilyte:
Could you remove unused functions?

Quasilyte:
Unfortunately, it does give mostly false positives on GOROOT.
We can try to handle `return ...` instead of `if ... {}` for simplicity,
but this will not solve all the problems.
These trigger a warning, but the nil check is not redundant there:
```go
lhs == nil || len(lhs) == 1
buf != nil && len(buf) == 0
```
Almost sure there are more cases.

Issue #300, opened 2018-06-22
lint: mark checkers with open issues as experimental


Quasilyte:
We've introduced "experimental" after many checkers were already implemented.

Proposing to mark problematic checkers as experimental and make them stable only after all known issues are fixed and they are properly tested on big Go projects.

Here is a list I've collected so far:
* docStub (see #261). Also, we need to improve heuristics.
* unnamedResult (see #259).
* unexportedCall. It's a complicated checker that received very few attention (testing-wise).
* stdExpr. Could be splitted to separate checkers and/or renamed.
* ptrToRefParam. Discussion is needed (see #302).

Quasilyte:
If anyone has more suggestions, we can re-open this.
Closing for now.

PR #301, opened 2018-06-22
lint: mark 4 checkers as experimental

Quasilyte:
docStub, stdExpr, unexportedCall and unnamedResult
are marked with experimental attirbute until they
are better tested and their bugs are fixed.

Updates #300


Quasilyte:
Now it's 5. There was only 4 in the 1st commit.
Final commit will be precise.

PR #310, opened 2018-06-23
cmd/criticize: replace GOPATH and GOROOT err loc prefix

Quasilyte:
```
/blah/path/to/goroot/src/pkg/file.go:1:2
=>
$GOROOT/src/pkg/file.go:1:1

Same for $GOPATH.
This behavior can be surpressed with `-shorterErrLocation=false` flag.
```

### DIFF
--- a/cmd/criticize/main.go
+++ b/cmd/criticize/main.go
@@ -245,11 +245,12 @@ func (l *linter) checkFile(f *ast.File) {
 }
 
 func shortenLocation(loc string) string {
-	if strings.HasPrefix(loc, build.Default.GOPATH) {
+	switch {
+	case strings.HasPrefix(loc, build.Default.GOPATH):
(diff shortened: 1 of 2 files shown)


cmd/criticize/main.go
```diff
@@ -229,10 +233,23 @@ func (l *linter) checkFile(f *ast.File) {
...
+				log.Printf("%s: %s: %v\n", loc, c.Rule, warn.Text)
 			}
 		}(c)
 	}
 	wg.Wait()
 }
+
+func shortenLocation(loc string) string {
+	if strings.HasPrefix(loc, build.Default.GOPATH) {
+		return strings.Replace(loc, build.Default.GOPATH, "$GOPATH", 1)
```
cristaloleg: Can be a switch:)

Quasilyte: Agreed.
Especially in a context that we're going to add more neat checks for switches.

PR #313, opened 2018-06-23
lint: add regexpMust checker

### LINKED ISSUE #209: lint: consider to use regexp.MustCompile instead of regexp.Compile
Name: `regexpChecker`
before:
```go
re, err := regexp.Compile("abc(d|e)") // const pattern
```
after:
```go
re := regexp.MustCompile("abc(d|e)")
```
From: https://github.com/cristaloleg/go-advices

Quasilyte:
At least for now, don't do same work as megacheck
and only suggest MustCompile|MustCompilePOSIX.

Fixes #209


Quasilyte:
Made typo in my nickname.. again. :[

PR #316, opened 2018-06-23
lint: add evalOrder checker

### LINKED ISSUE #73: lint: add "dangerous depencency on evaluation order" checker
Some expressions and statements in Go may evaluate differently on various implementations due to unspecified evaluation order details.

See https://github.com/Quasilyte/goism/issues/115 for more info.

Quasilyte:
Needed IsTypeExpr, so moved its implementation from
astwalk to a new package lintutil, so both lint and
astwalk can use it.

As 2 checkers now use AstSet-like type, moved it to lintutil.

Fixes #73


Quasilyte:
Not very powerful and does sometimes report false positives (it does not verify that function that takes a pointer really modifies a value).
Experimental until this issue is resolved.

lint/internal/lintutil/ast_set.go
```diff
@@ -0,0 +1,44 @@
+package lintutil
```
cristaloleg: WDYT s/`ast_set.go`/`astset.go` ? or even `set.go`

lint/evalOrder_checker.go
```diff
@@ -0,0 +1,156 @@
...
+
+func (c *evalOrderChecker) VisitStmt(stmt ast.Stmt) {
+	switch stmt := stmt.(type) {
+	case *ast.ReturnStmt:
+		c.checkReturn(stmt)
+	case *ast.AssignStmt:
+		// TODO(quasilyte):
+		//	https://github.com/golang/go/issues/23188
+		//	https://github.com/golang/go/issues/24448
+		//	https://github.com/golang/go/issues/23017
```
cristaloleg: + https://github.com/golang/go/issues/25609 ?

Quasilyte: It's referenced inside test files.

lint/internal/lintutil/ast_set.go
Quasilyte: Fixed.

lint/evalOrder_checker.go
```diff
@@ -0,0 +1,156 @@
...
+		return
+	}
+	c.reset()
+	for _, x := range ret.Results {
+		c.collectDeps(x)
+	}
+	// TODO(quasilyte): we can parametrize the threshold later.
+	if deps := c.depsCount(); deps != 0 {
+		c.warn(ret, deps)
+		return
```
Quasilyte: This return is redundant.

PR #319, opened 2018-06-23
lint/yodaStyleExpr

### LINKED ISSUE #292: line/yodaStyleExpr: do not place const value as a first operand in binary expression
Before:
```go
if nil != ptr { //
```
After:
```go
if ptr != nil { //
```

cristaloleg:
Fixes #292

### DIFF
--- a/cmd/gocritic/testdata/yodaStyleExpr/positive_tests.go
+++ b/cmd/gocritic/testdata/yodaStyleExpr/positive_tests.go
@@ -2,17 +2,17 @@ package checker_test
 
 func f1() {
 	var m map[int]int
-	/// consider to change order of expression in nil == m
+	/// consider to change order in expression to m == nil
 	if nil == m {
 	}
 
 	var a int
-	/// consider to change order of expression in 10 == a
+	/// consider to change order in expression to a == 10
 	if 10 == a {
 	}
 
 	var s string
-	/// consider to change order of expression in "" == s
+	/// consider to change order in expression to s == ""
 	if "" == s {
 	}
 }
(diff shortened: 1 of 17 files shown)


cmd/gocritic/testdata/yodaStyleExpr/positive_tests.go
```diff
@@ -0,0 +1,47 @@
...
+	switch {
+	/// consider to change order of expression in nil == ch
+	case nil == ch:
+		//
+	}
+	/// consider to change order of expression in nil == ch
+	return nil == ch
+}
+
+type kek struct {
```
Quasilyte: Can we avoid such type names, please?

cmd/gocritic/testdata/yodaStyleExpr/positive_tests.go
```diff
@@ -0,0 +1,47 @@
...
+	var k kek
+	/// consider to change order of expression in 0 == k.a
+	if 0 == k.a {
+	}
+}
+
+func f4() {
+	var a int
+	f := func(bool) {}
+	/// consider to change order of expression in 10 == a
```
Quasilyte: You can do `astcopy` of triggered binary expr, swap its X and Y, and then suggest it as a replacement.

lint/yodaStyleExpr_checker.go
```diff
@@ -0,0 +1,55 @@
...
+	checkerBase
+}
+
+func (c *yodaStyleExprChecker) VisitLocalExpr(expr ast.Expr) {
+	binexpr, ok := expr.(*ast.BinaryExpr)
+	if !ok {
+		return
+	}
+	if binexpr.Op == token.EQL || binexpr.Op == token.NEQ {
+		if c.isConstExpr(binexpr.X) && c.isVar(binexpr.Y) {
```
Quasilyte: Why not `c.isConstExpr(binexpr.X) && !c.isConstExpr(binexpr.Y)`?
Would it break something?

Quasilyte: Or is it a precaution to avoid suggestions in a potentially long (multi-line) expressions?
Would make sense.. But would give less valid warnings as well.

lint/yodaStyleExpr_checker.go
```diff
@@ -0,0 +1,55 @@
...
+		}
+	}
+}
+
+func (c *yodaStyleExprChecker) isConstExpr(expr ast.Expr) bool {
+	if qualifiedName(expr) == "nil" {
+		return true
+	}
+	_, ok := expr.(*ast.BasicLit)
+	return ok
```
Quasilyte: `return qualifiedName(expr) == "nil" || astp.IsBasicLit(expr)`?

Issue #296, opened 2018-06-22
cmd/makedocs: add tests for generated output validation


quasilyte:
It's tedious to check `overview.md` manually.

At the very least we should test that documentation comment is parsed correctly, without extra
whitespace and other artifacts.

It would be also good to test for expected errors to make sure invalid doc comments are rejected.

Quasilyte:
We need to discuss alternative approach for documentation proposed in #322.

Quasilyte:
CC @cristaloleg, @fexolm

cristaloleg:
> without extra whitespace and other artifacts.
is this soooo critical? 😉 

I understand - perfect documentation is sweet, but having small and minor problems is ok.
At least we have docs and they're up to date.

Having test for docs is an overkill.

Quasilyte:
This is not about that.
Please take a look at what PR does.

PR #322, opened 2018-06-23
cmd/makedocs: add tests for generated output validation

weeellz:
#296


cmd/makedocs/check-docs_test.go
```diff
@@ -0,0 +1,76 @@
...
+		Before:           "Single line for code",
+		After:            "After issue is fixed",
+	}
+
+	testFiles := parseFiles()
+
+	for _, file := range testFiles {
+		if file.ShortDescription == wantDoc.ShortDescription {
+			continue
+		} else {
```
cristaloleg: unneeded :)

cmd/makedocs/check-docs_test.go
```diff
@@ -0,0 +1,76 @@
...
+			t.Errorf("%s: ShortDescription mismatch:\nhave: %q\nwant: %q", file.Name,
+				file.ShortDescription, wantDoc.ShortDescription)
+		}
+		if file.Before == wantDoc.Before {
+			continue
+		} else {
```
cristaloleg: also unneeded

cmd/makedocs/check-docs_test.go
```diff
@@ -0,0 +1,76 @@
...
+			t.Errorf("%s: @Before mismatch:\nhave: %q\nwant: %q", file.Name,
+				file.Before, wantDoc.Before)
+		}
+		if file.After == wantDoc.After {
+			continue
+		} else {
```
cristaloleg: and again :)

cmd/makedocs/testdata/bar.go
```diff
@@ -0,0 +1,9 @@
+package lint
+
+//! This is short desc kek
```
cristaloleg: There was a proposal to omit kek/lol/lel to more classic foo/bar/baz 😉

cmd/makedocs/check-docs_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+func parseFiles() []checkerDoc {
+	tests, err := ioutil.ReadDir("./testdata")
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	pkgs, err := parser.ParseDir(&token.FileSet{}, "./testdata",
+		func(inf os.FileInfo) bool {
```
cristaloleg: Can be be extracted to a variable `filter`, will be a bit simpler to read the code.

cmd/makedocs/check-docs_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"io/ioutil"
+	"log"
+	"os"
+	"strings"
+	"testing"
+)
+
+var docCheckers []checkerDoc
+
+func parseFiles() []checkerDoc {
```
Quasilyte: I though you would use same function as `main.go`.
Maybe `parseFiles` can accept path argument?
Please, move it to main.go and use it in both places.

weeellz: > I though you would use same function as `main.go.`
What function?

Quasilyte: One I though you would write to avoid code duplication between test and main.go.
If we're testing main.go, it should use same function as the main.go

cmd/makedocs/check-docs_test.go
fexolm: @cristaloleg, @Quasilyte, we could add checker for this.
**before:**
```go
if cond {
     continue
} else {
     ...
}
```
**after:**
```go
if !cond {
    ...
}
```

cmd/makedocs/main.go
```diff
@@ -160,8 +152,41 @@ func parseNote(lines []string, index *int, c *checker) error {
...
+func parseComments(file *ast.File, cd *checkerDoc) {
+	for _, comment := range file.Comments {
+		if strings.HasPrefix(comment.Text(), "!") {
+			parseComment(comment.Text(), cd)
+			break
+		}
+	}
+}
+
+func parseFiles(path string) []checkerDoc {
```
cristaloleg: unused, that's why you've a failing build

weeellz:
```
gometalinter.v2 --skip=testdata --vendor ./...
cmd/makedocs/main.go:184:1:warning: parseFiles is unused (deadcode)
```
Hmmm

cmd/makedocs/main.go
weeellz: TY.

weeellz:
Fails intentionally with incorrect doc.

Quasilyte:
I'll take a closer look closer to a day time. :)

Quasilyte:
I'm going to merge this and suggest fixes in another PR.

Quasilyte:
OK, I have an idea. Will implement it today (read: later).

Quasilyte:
The idea is to avoid parsing, so no parsing tests are needed.
I think the loss in convenience is low enough to give it a try.
For single line before/after it even looks better.

Quasilyte:
@fexolm, could you approve this one, if there is no objections against it?

Quasilyte:
@cristaloleg, PTAL.
If you approve this one, I'll merge it.

PR #323, opened 2018-06-24
lint/emptyFmt

### LINKED ISSUE #107: lint: add empty-fmt checker
For some reasons, this kind of code exists in our own codebase:
```go
x := fmt.Sprintf("abc")
```
It uses formatted print function, but does not specify and formatting arguments, neither string itself has any formatting directives.
If there are actually missing arguments for format directives, go vet gives a warning:
```go
x := fmt.Sprintf("%d")
// Sprintf format %d reads arg #1, but call has 0 args (vet)
```

So, our task is to check formatted IO calls for arguments count.

For `fmt.Sprintf` it's a bad practice to use 1 argument.
For `fmt.Fprintf` it's a bad practice to use 2 arguments (because 1st argument is a file + formatting string itself).
And so on. There are many such functions in go, `log` package is a second lead.

This checker should be implemented using `baseExprChecker`.

cristaloleg:
Fixes #107

### DIFF
--- a/lint/emptyFmt_checker.go
+++ b/lint/emptyFmt_checker.go
@@ -4,11 +4,11 @@ package lint
 //
 // @Before:
 // fmt.Sprintf("whatever")
-// fmt.Errorf("whereever")
+// fmt.Errorf("wherever")
 //
 // @After:
 // fmt.Sprint("whatever")
-// errors.New("whereever")
+// errors.New("wherever")
 
 import (
 	"go/ast"
--- a/lint/emptyFmt_checker.go
+++ b/lint/emptyFmt_checker.go
@@ -1,0 +1,57 @@
+package lint
+
+//! Detects usages of formatting functions without formatting arguments.
+//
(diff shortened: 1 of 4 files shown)


lint/emptyFmt_checker.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+	case 2:
+		if name == "fmt.Fprintf" {
+			c.warn(call, name, "fmt.Fprint")
+		}
+	}
+}
+
+func (c *emptyFmtChecker) warn(cause *ast.CallExpr, original, suggestion string) {
+	c.ctx.Warn(cause, "consider to change function to %s", strings.Replace(original, original, suggestion, 1))
```
Quasilyte: Why not just `suggestion` instead of `strings.Replace(original, original, suggestion, 1)`?

cristaloleg: Well, yeah ;D

PR #324, opened 2018-06-24
lint/docStub: do not check unexported funcs, warn about XXX

### LINKED ISSUE #261: lint: docStub false positives

cristaloleg:
Fixes #261

### DIFF
--- a/cmd/gocritic/testdata/docStub/negative_tests.go
+++ b/cmd/gocritic/testdata/docStub/negative_tests.go
@@ -1,6 +1,6 @@
 package checker_test
 
 // Bar returns string.
-func Bar() {
+func Bar() {}
 
-}
+func foo() {}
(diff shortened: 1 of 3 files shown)


Quasilyte:
We may also want to check whether there is only 1 comment line.
If there is more than 1, it's not a stub.

PR #326, opened 2018-06-24
lint: add appendAssign checker

Quasilyte:
Detects suspicious assignments of append results.
Does recognize several popular slice/append idioms
to reduce the amount of false positives.

This kind of error may occur when append assignment being copy/pasted.

### DIFF
--- a/cmd/gocritic/testdata/appendAssign/negative_tests.go
+++ b/cmd/gocritic/testdata/appendAssign/negative_tests.go
@@ -1,0 +1,85 @@
+package checker_tests
+
+func normalAppends() {
+	var xs, ys []int
+
+	xs = append(xs, 1)
+	ys = append(ys, 1, 2)
+	xs = append(xs, ys[0], xs[0])
+}
+
+func permittedAppends() {
+	var xs, ys []int
+
+	// We're trying to detect `x = append(y, ...)` patterns
+	// where y is used instead of x by mistake, so lines below
+	// do not trigger a warning.
+
+	xs0 := append(xs, 1)
+	xs1 := append(xs, 1)
+	ys0 := append(ys, 1)
+
+	// Also permit to assign to "_".
+	_ = append(xs, xs0[0], xs1[1], ys0[0])
+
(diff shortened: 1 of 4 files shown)


docs/overview.md
```diff
@@ -395,14 +422,14 @@ Detects usages of formatting functions without formatting arguments.
 **Before:**
 ```go
 fmt.Sprintf("whatever")
-fmt.Errorf("whereever")
+fmt.Errorf("wherever")
```
Quasilyte: Happened due to docs re-generation.
And I think there is a typo: should be `whatever`, not `wherever`. Right?

lint/appendAssign_checker.go
```diff
@@ -0,0 +1,100 @@
...
+	for i, rhs := range assign.Rhs {
+		call, ok := rhs.(*ast.CallExpr)
+		if !ok || qualifiedName(call.Fun) != "append" {
+			continue
+		}
+		c.checkAppend(assign.Lhs[i], call)
+	}
+}
+
+func (c *appendAssignChecker) matchSlices(cause ast.Node, x, y ast.Expr) {
```
cristaloleg: IMO, can we move this method after `checkAppend`? :3

cmd/gocritic/testdata/appendAssign/negative_tests.go
```diff
@@ -0,0 +1,85 @@
...
+
+func appendNotInAssignment() {
+	var xs, ys []int
+
+	// These are somewhat weird, but has nothing
+	// to do with diagnostic this checker wants to perform.
+
+	var v1 = append(xs, 1)
+	var (
+		_  = 1
```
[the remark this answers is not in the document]
Quasilyte: Removed now.
Used it to skip 1st spec, but there is no big reason for that, doesn't improve coverage.

lint/appendAssign_checker.go
Quasilyte: Done.

PR #329, opened 2018-06-24
lint: add namedConst checker

### LINKED ISSUE #215: lint: add enumValue checker
Detects when literal value is coerced to enum-like type that contains a name for used value.
Suggests using that named constant instead of raw literal.

**Before:**
```go
type color int
const (
  colorDefault color = iota
  colorRed
)
func f() color { return 0 }
```
**After:**
```go
type color int
const (
  colorDefault color = iota
  colorRed
)
func f() color { return colorDefault }
```

Quasilyte:
Detects raw literal values that can be replaced
by already defined (and imported) named constant.

Fixes #215

### DIFF
--- a/cmd/gocritic/testdata/namedConst/negative_tests.go
+++ b/cmd/gocritic/testdata/namedConst/negative_tests.go
@@ -1,0 +1,25 @@
+package checker_tests
+
+import (
+	"time"
+)
+
+func returnExpr() {
+	// The value is 2 (colorGreen), but it's an expression,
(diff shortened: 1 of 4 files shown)


Quasilyte:
There is a subtle issue with submitted implementation.
It does trigger on definitions as well as usages.

This simple code will generate 2 warnings:
```go
type typename string

const (
	tnInt    typename = "int"
	tnString typename = "string"
)
```
It will suggest to use `tnInt` and `tnString` instead of literals.
Should be fixed before this is merged.

PR #335, opened 2018-06-25
all: version command

### LINKED ISSUE #186: cmd/gocritic: add -v flag to show current version.
We should also start versioning our project.

cristaloleg:
Fixes #186

### DIFF
--- a/cmd/gocritic/main.go
+++ b/cmd/gocritic/main.go
@@ -39,7 +39,7 @@ var subCommands = []*subCommand{
 	{
 		main:  printVersion,
 		name:  "version",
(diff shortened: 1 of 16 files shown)


cmd/gocritic/main.go
```diff
@@ -61,6 +68,10 @@ func main() {
 	cmd.main()
 }
 
+func printVersion() {
+	fmt.Printf("go-critic version: %s\n", version)
```
Quasilyte: Could you use `log.Printf` here?
For consistency with other sub commands.

It will output `version: ` prefix, so you can just do `log.Println(version)`, for example.

PR #336, opened 2018-06-26
Hotfix/304

### LINKED ISSUE #304: lint: importShadow handling of renamed imports
For packages that are imported with renaming, `importShadow` gives no warnings

If it's expected behavior, negative tests should be added.
If it should trigger on them, fix should be attached to this issue.

Reproducer:
```go
package checker_test

import (
	xmath "math"
)

var _ = xmath.E // Make sure xmath is used

//TODO: shadow of imported package 'xmath'
func renamedImportShadow(xmath int) {}
```

NexoMichael:
Fixes #304

### DIFF
--- a/lint/importShadow_checker.go
+++ b/lint/importShadow_checker.go
@@ -42,18 +42,27 @@ type importShadowChecker struct {
 }
 
 func (c *importShadowChecker) VisitLocalDef(def astwalk.Name, _ ast.Expr) {
-	for _, imp := range c.ctx.Context.pkg.Imports() {
-		if imp.Name() == def.ID.Name {
-			c.warnImportShadowed(def.ID, imp)
+	for _, v := range c.ctx.Context.typesInfo.Defs {
+		if v == nil || v.Parent() == nil {
+			continue
+		}
+		for _, name := range v.Parent().Names() {
+			pkg, ok := v.Parent().Lookup(name).(*types.PkgName)
+			if !ok {
+				continue
+			}
+			if name == def.ID.Name {
+				c.warnImportShadowed(def.ID, name, pkg.Imported())
+			}
 		}
 	}
 }
 
-func (c *importShadowChecker) warnImportShadowed(id ast.Node, pkg *types.Package) {
+func (c *importShadowChecker) warnImportShadowed(id ast.Node, importedName string, pkg *types.Package) {
 	if pkg.Path() == pkg.Name() {
 		// check for standart library packages
-		c.ctx.Warn(id, "shadow of imported package '%s'", pkg.Name())
+		c.ctx.Warn(id, "shadow of imported package '%s'", importedName)
 	} else {
-		c.ctx.Warn(id, "shadow of imported from '%s' package '%s'", pkg.Path(), pkg.Name())
+		c.ctx.Warn(id, "shadow of imported from '%s' package '%s'", pkg.Path(), importedName)
 	}
 }
--- a/lint/importShadow_checker.go
+++ b/lint/importShadow_checker.go
@@ -46,15 +46,15 @@ func (c *importShadowChecker) VisitLocalDef(def astwalk.Name, _ ast.Expr) {
 		if v == nil || v.Parent() == nil {
 			continue
 		}
-		for _, name := range v.Parent().Names() {
-			pkg, ok := v.Parent().Lookup(name).(*types.PkgName)
-			if !ok {
-				continue
-			}
-			if name == def.ID.Name {
-				c.warnImportShadowed(def.ID, name, pkg.Imported())
-			}
+		elem := v.Parent().Lookup(def.ID.Name)
+		if elem == nil {
+			continue
 		}
+		pkg, ok := elem.(*types.PkgName)
+		if !ok {
+			continue
+		}
+		c.warnImportShadowed(def.ID, def.ID.Name, pkg.Imported())
 	}
 }
 
--- a/cmd/gocritic/testdata/importShadow/positive_tests.go
+++ b/cmd/gocritic/testdata/importShadow/positive_tests.go
@@ -47,5 +47,5 @@ type shadower struct{}
 /// shadow of imported package 'fmt'
 func (fmt shadower) f() {}
 
-//TODO: shadow of imported package 'xmath'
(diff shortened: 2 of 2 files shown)


lint/importShadow_checker.go
```diff
@@ -42,18 +42,27 @@ type importShadowChecker struct {
 }
 
 func (c *importShadowChecker) VisitLocalDef(def astwalk.Name, _ ast.Expr) {
-	for _, imp := range c.ctx.Context.pkg.Imports() {
-		if imp.Name() == def.ID.Name {
-			c.warnImportShadowed(def.ID, imp)
+	for _, v := range c.ctx.Context.typesInfo.Defs {
```
Quasilyte: I'm a bit concerned that it iterated all definitions for every local definition.
This can be noticeable when checking big projects like 
`gocritic check-project -enable=importShadow $GOROOT/src`
I think it will take measurable amount of time (we have `unusedParam` that works several minutes vs 1-2secs of other checkers).
I'll think about this after working hours, maybe there is a better 
We can still merge this solution, but it would be also good to either add TODO notice there or create an issue that this checker may be slow and could benefit from tuning.

NexoMichael: Completely agree, currently it is O(n^2) for count of definitions - and it is really bad.

Maybe we should add some "precomputation" stage, for example it could look like - each linter implements some function that is called for example for each package or for each file, during this stage we will calculate all supplemental maps/structures which then will be used in the real linter checkers.

Quasilyte: This is 3rd checker that can benefit from precomputed data.
The only issue is that it's not optimal to precompute all such data without knowing it's 100% needed, so we need to know what kind of resources checker needs, and if specific indexes has no usages (because related checkers are not enabled during the run), we should not load them.

For now, it can be simpler to track all such linters and make a solution that fits them all later.

I would also like to think about this particular check. We may be lacking some knowledge that could make O(n^2) not needed at all, even without precomputed map.

(If @cristaloleg approves, we may merge as is and improve on it later.)

lint/importShadow_checker.go
```diff
@@ -42,18 +42,27 @@ type importShadowChecker struct {
...
+		if v == nil || v.Parent() == nil {
+			continue
+		}
+		for _, name := range v.Parent().Names() {
```
Quasilyte: AFAIK `Names()` call does build slice for every call. It also does sorting.
Not very good API for this use case. :(

lint/importShadow_checker.go
NexoMichael: I have some proposal for this precomputations it looks like this:
1) We define some Option flags for the linters, like PrefetchImportedPackages, etc...
2) During linter registration it can define as many options as it needs, options could be file or package specific
3) Before real running linters in the goroutines, all precomputations needed are summarized and done which results in more filled context
4) linters use precomputed data from the context

It is also possible to "hack"-fix current approach, we can do this by doing not real caching, for example - in the beginning of the linter check we check local cache object if current file was already precomputed, and if so use this precomputed (in my case - map[file path]map[package name]bool structure)

From my perspective second approach is not very good cause precomputed data can't be shared between different linters, but for one-point-optimization it looks ok.

lint/importShadow_checker.go
NexoMichael: yes it is... it is used here because (*types.Scope).elems is not exported and only possible solution to iterate over it is to use (*types.Scope).Lookup(name string), let me remove this iteration completely so it become a little bit faster.

NexoMichael: done

NexoMichael:
Here is proposal for one-point fix

```
import (
	"go/ast"
	"go/types"
	"sync"

	"github.com/go-critic/go-critic/lint/internal/astwalk"
)

func init() {
	addChecker(&importShadowChecker{}, attrExperimental)
}

type importShadowChecker struct {
	checkerBase

	// stores for each "filename" information about all imported packages
	// map[file name]map[imported package name]package object
	imports     map[string]map[string]*types.Package
	importsLock sync.Mutex
}

func (c *importShadowChecker) VisitLocalDef(def astwalk.Name, _ ast.Expr) {
	pkg, found := c.getDefinedPackages()[def.ID.Name]
	if found && pkg != nil {
		c.warnImportShadowed(def.ID, def.ID.Name, pkg)
	}
}

func (c *importShadowChecker) warnImportShadowed(id ast.Node, importedName string, pkg *types.Package) {
	if pkg.Path() == pkg.Name() {
		// check for standart library packages
		c.ctx.Warn(id, "shadow of imported package '%s'", importedName)
	} else {
		c.ctx.Warn(id, "shadow of imported from '%s' package '%s'", pkg.Path(), importedName)
	}
}

func (c *importShadowChecker) getDefinedPackages() map[string]*types.Package {
	c.importsLock.Lock()
	defer c.importsLock.Unlock()

	if c.imports == nil {
		c.imports = map[string]map[string]*types.Package{}
	}

	imports, found := c.imports[c.ctx.filename]
	if found {
		return imports // fast path
	}

	imports = map[string]*types.Package{}
	for _, v := range c.ctx.Context.typesInfo.Defs {
		if v == nil || v.Parent() == nil {
			continue
		}
		for _, name := range v.Parent().Names() {
			elem := v.Parent().Lookup(name)
			if elem == nil {
				continue
			}
			pkg, ok := elem.(*types.PkgName)
			if !ok {
				continue
			}
			imports[name] = pkg.Imported()
		}
	}

	c.imports[c.ctx.filename] = imports
	return imports
}
```

What do you think?

cristaloleg:
@NexoMichael thank you 😉

Quasilyte:
> What do you think?

I can comment that mutex is not needed since embedding app should not use single checker instance between multiple goroutines. `cmd/criticize` runs checkers in parallel, but this is N-1 concurrency. If application needs N-M, it should spawn M checkers (this is why `addChecker` does create a prototype and clones it inside `NewChecker`).

Regarding performance fix: I'll take a look later. I don't have exact run stats to evaluate how big the problem is and by how much this change would help it. (Can't start looking into it until evening.)

Quasilyte:
This property makes checkers somewhat easier as there is no concerns about multi-entry for a single instance.

PR #340, opened 2018-06-26
all: move test to lint package

cristaloleg:
So...move tests from `cmd/gocritic` to `lint`.
From now we'll be able to check our coverage(it's very good 94.4%)
In next PR I'm going to enable automatic coverage tool(https://codecov.io/ objections?)
Also I've removed `///` comments in `negative_tests.go` to simplify checking regexp. I think it's not so important.
And on more: I left tests in `cmd/gocritic` until we'll be 100% sure work is done. Maybe I will remove them even in this PR.


Makefile
```diff
@@ -15,5 +15,5 @@ docs:
 ci:
 	go get -t -v ./...
 	go tool vet .
-	go test -v -race ./...
+	go test -v -race -count=1 ./...
```
cristaloleg: don't cache tests

lint/lint_test.go
```diff
@@ -0,0 +1,169 @@
...
+		for _, w := range f.warnings[line] {
+			if w.matched {
+				continue
+			}
+			t.Errorf("%s:%d: unmatched `%s`", testFilepath, line, w)
+		}
+	}
+}
+
+func getWarnLine(ctx *Context, node ast.Node) int {
```
cristaloleg: We might be interested into including warn's line in `Warning`.

lint/lint_test.go
```diff
@@ -0,0 +1,169 @@
...
+	loc := ctx.FileSet().Position(node.Pos()).String()
+	num, _ := strconv.Atoi(strings.Split(loc, ":")[1])
+	return num
+}
+
+func getFilename(prog *loader.Program, f *ast.File) string {
```
cristaloleg: We might omit this method somehow too.

lint/lint_test.go
Quasilyte: Isn't it trivial to get line from `Warning` if you have `token.FileSet`?

lint/lint_test.go
```diff
@@ -0,0 +1,169 @@
...
+	// see https://github.com/golang/go/issues/24498
+	fname := prog.Fset.Position(f.Pos()).String() // ex: /usr/go/src/pkg/main.go:1:1
+	fname = strings.Split(fname, ":")[0]          // ex: /usr/go/src/pkg/main.go
+	return filepath.Base(fname)                   // ex: main.go
```
Quasilyte: Can be simplified to:
```go
filepath.Base(prog.Fset.Position(f.Pos()).Filename)
```

PR #343, opened 2018-06-26
lint: add commentedOutCode checker

### LINKED ISSUE #337: lint: add commented-out code detector
Detects commented-out code inside function bodies.

**Before:**
```go
// fmt.Println("Debugging hard")
foo(1, 2)
```
**After:**
```go
foo(1, 2)
```

Not sure this can be done without both false positives and false negatives, but still would be useful if it can detect at least some instances. Can help during big and old (legacy) codebase cleanup.

Quasilyte:
Fixes #337


lint/commentedOutCode_checker.go
```diff
@@ -0,0 +1,101 @@
+package lint
+
+//! Detects commented-out code inside function bodies.
+//
+// @Before:
+// // fmt.Println("Debugging hard")
```
cristaloleg: `is hard` ? :)

lint/internal/astwalk/local_comment_visitor.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+func (w *localCommentVisitor) WalkFile(f *ast.File) {
+	for _, decl := range f.Decls {
+		decl, ok := decl.(*ast.FuncDecl)
+		if !ok || !w.visitor.EnterFunc(decl) {
+			continue
+		}
+		begin := decl.Pos()
+		end := decl.End()
```
cristaloleg: `begin, end := decl.Pos(), decl.End()` :3

lint/commentedOutCode_checker.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (c *commentedOutCodeChecker) isPermittedExpr(x ast.Expr) bool {
+	// Permit anything except expressions that can be used
+	// with complete result discarding.
+	switch x := x.(type) {
+	case *ast.CallExpr:
+		return false
+	case *ast.UnaryExpr:
+		// "<-" channel receive is not permitted.
```
cristaloleg: But why?

lint/commentedOutCode_checker.go
Quasilyte: It's like `Party hard`.

lint/commentedOutCode_checker.go
Quasilyte: Here is explanation (may include into the code as a comment):
We start scanning from statements. If there is `ExprStmt`, we need
to determine whether it looks like commented-out code or not.
To do that, we use Go compiler property that forbids things like binary expr without using it's results. So, `// a + b` does not cause any warnings because it's `*ast.BinaryExpr`, and it's a good thing, because if we remove `//` this code would not compile, you can't use `a + b` without assigning it to something => so it's not commented-out code but some kind of comment that just happens looks like Go code.
Not so with function calls and recv expressions like `<- ch`.
They can be in place of statement without assignment (for side effects).
This means they could represent commented-out code.

lint/internal/astwalk/local_comment_visitor.go
Quasilyte: Removed tmp vars.

lint/commentedOutCode_checker.go
cristaloleg: That's cool. It would be good to extend comment, but with a shorter addition, like: <- is permitted, 'cause `<-ch` can be an independent expression.

Quasilyte: Few lines above:
```go
	// Permit anything except expressions that can be used
	// with complete result discarding.
```

cristaloleg: kk

Quasilyte:
PTAL.

Quasilyte:
Um. Seems like I've removed sanity checks.

cristaloleg:
Have restarted a build due timeout 😉

Issue #352, opened 2018-06-27
all: perf issues


cristaloleg:
Some of the checkers consume more time than they should(citation needed).
We might be interested into speeding them:
```bash
    --- PASS: TestCheckers/emptyFmt (0.84s)
    --- PASS: TestCheckers/flagDeref (0.66s)
    --- PASS: TestCheckers/importShadow (1.90s)
    --- PASS: TestCheckers/namedConst (0.40s)
    --- PASS: TestCheckers/rangeValCopy (0.76s)
    --- PASS: TestCheckers/regexpMust (0.59s)
```

Quasilyte:
Maybe due to imports?
If you're talking about tests.
This is why we avoid imports in testdata whether possible.

cristaloleg:
Yep, looks like it's due imports and most of this time was consumed by loader. Closing.

PR #357, opened 2018-06-28
lint/indexOnlyLoop: add index-only range loops checker

### LINKED ISSUE #72: lint: add index-only range loops checker
If range loop is iterating over `container` which has pointer-type elements (that is, any `*T` or reference-like type as `map` or `chan`) using only `key` variable (no `value`) and `container[key]` occurs more than once inside loop body, suggest using `for key, v := range container` form.

This code triggers a warning:
```go
func closeFiles(files []*os.File) {
	for i := range files {
		if files[i] != nil {
			files[i].Close()
		}
	}
}
```
This is suggested form:
```go
func closeFilesSuggester(files []*os.File) {
	for _, f := range files {
		if f != nil {
			f.Close()
		}
	}
}
```

NexoMichael:
Fixes #72


lint/testdata/indexOnlyLoop/negative_tests.go
```diff
@@ -0,0 +1,19 @@
...
+		fmt.Println(i)
+	}
+}
+
+func closeNonPtrFiles(files []os.File) {
+	for i := range files {
+		fmt.Println(files[i].Name())
+		files[i].Close()
+	}
+}
```
fexolm: please add test where index used with more that one array, like
```go
for i := range files {
    a[i] = b[i]
}
```

Quasilyte:
I'll take a look in a few (10-20) minutes.

lint/indexOnlyLoop_checker.go
```diff
@@ -0,0 +1,81 @@
+package lint
+
+import (
+	"fmt"
+	"go/ast"
+)
+
+//! Detects when range loop is iterating over container which has pointer-type elements (that is, any *T or reference-like type as map or chan) using only key variable (no value) and container[key] occurs more than once inside loop body
```
Quasilyte: Can we make summary shorter?
Maybe something like this is concise enough:
`Detects for loops that can benefit from rewrite to range loop.`

lint/testdata/indexOnlyLoop/negative_tests.go
```diff
@@ -0,0 +1,27 @@
...
+
+func indexReuse(filesA []*os.File, filesB []*os.File) {
+	for i := range filesA {
+		if filesA[i] == filesB[i] {
+			fmt.Println("equal")
```
Quasilyte: We generally avoid unnecessary imports in the tests because they cause significant slowdown (until we reuse loader program between tests).
For example, 0 imports => 0.0s test time, 1-2 imports => 1s test time.
This effects sums up.
See also #352.
This is not requirement, just a reminder.

lint/indexOnlyLoop_checker.go
```diff
@@ -0,0 +1,81 @@
...
+	addChecker(&indexOnlyLoopChecker{}, attrExperimental)
+}
+
+type indexOnlyLoopChecker struct {
+	checkerBase
+}
+
+func (c *indexOnlyLoopChecker) VisitStmt(stmt ast.Stmt) {
+	if s, ok := stmt.(*ast.RangeStmt); ok && s.Key != nil {
+		sX := s.X.(*ast.Ident).Obj
```
Quasilyte: This will panic for things like `for i := range x.slice` or anything except `*ast.Ident`.
Should probably use `sX, ok :=` here, to avoid that.

Quasilyte: Also, `ast.Object` is quite deprecated thing. You may want to use `types.Object`.
(Okay for first impl, just a notice.)

lint/indexOnlyLoop_checker.go
```diff
@@ -0,0 +1,81 @@
...
+		var tp ast.Expr
+		switch sXField := sX.Decl.(type) {
```
Quasilyte: `s/sXField/decl/`? (optional.)

lint/indexOnlyLoop_checker.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+type indexOnlyLoopChecker struct {
+	checkerBase
+}
+
+func (c *indexOnlyLoopChecker) VisitStmt(stmt ast.Stmt) {
+	if s, ok := stmt.(*ast.RangeStmt); ok && s.Key != nil {
+		sX := s.X.(*ast.Ident).Obj
+		var tp ast.Expr
```
Quasilyte: Regarding naming things: `typ` is more usual shortcut for "type". (At least in my experience.)
(Can leave as is. This is only a notice.)

lint/indexOnlyLoop_checker.go
```diff
@@ -0,0 +1,81 @@
...
+		case *ast.Field: // go 1.10
+			tp = sXField.Type
+		case *ast.ValueSpec: // go 1.11
+			tp = sXField.Type
+		default:
+			return
+		}
+		// sX should always be of *ast.ArrayType type
+		// cause we are in *ast.RangeStmt statement
+		sxFiledType, ok := tp.(*ast.ArrayType)
```
Quasilyte: You could use `*types.ArrayType`/`*types.SliceType` to get same info.

lint/indexOnlyLoop_checker.go
```diff
@@ -0,0 +1,81 @@
...
+		count := 0
+		ast.Inspect(stmt, func(n ast.Node) bool {
+			if iExpr, ok := n.(*ast.IndexExpr); ok {
+				x := iExpr.X.(*ast.Ident).Obj
+				key := iExpr.Index.(*ast.Ident).Obj
+				if x == sX && key == sKey {
+					count++
+				}
+			}
+			// stop DFS traverse if we found more then one usage
```
Quasilyte: `s/more then/more than/`

Quasilyte:
Thanks for contributing!
Is it more convenient for you to fix issues in another PR or inside this one?
We can go either way.

NexoMichael:
> Is it more convenient for you to fix issues in another PR or inside this one?

I'll fix in this PR.

lint/indexOnlyLoop_checker.go
NexoMichael: fixed

lint/testdata/indexOnlyLoop/negative_tests.go
NexoMichael: removed imports

lint/indexOnlyLoop_checker.go
NexoMichael: 1. fixed panics on slices, added tests for that
2. don't quite understand how I can replcae *ast.Object with types Object interface here...

lint/indexOnlyLoop_checker.go
NexoMichael: done

lint/indexOnlyLoop_checker.go
NexoMichael: seems that I don't get what you mean here...

lint/indexOnlyLoop_checker.go
NexoMichael: added more detailed comment, I did't find way in current impl how `*types.SliceType` could arrive here... have some thoughts what test can be for that?

lint/indexOnlyLoop_checker.go
cristaloleg: rename `sXField` to `decl` 😉

NexoMichael: already renamed to `sXFieldDecl`... I got it, it was simple sed...

lint/indexOnlyLoop_checker.go
Quasilyte: This suggestion was for `types.Object`.
Let's merge this versions and I'll see if it can be done without `ast.Object` later.

PR #358, opened 2018-06-28
lint/deferInLoop: initial implementation

### LINKED ISSUE #338: lint/deferInLoop:
Name: `deferInLoop`
Before:
```go
for i := range [10]int{} {
	defer f(i) // will be executed only at the end of func
}
```
After:
```go
for i := range [10]int{} {
	func() {
		defer f(i) 
	}
}
```
See: https://play.golang.org/p/UjlP3nnRHkT

cristaloleg:
Fixes #338

### DIFF
--- a/lint/deferInLoop_checker.go
+++ b/lint/deferInLoop_checker.go
@@ -1,0 +1,51 @@
+package lint
+
+//! Detects defer in loop and warns that it will not be executed till the end of function's scope.
+//
+// @Before:
+// for i := range [10]int{} {
+// 		defer f(i) // will be executed only at the end of func
+// }
+//
+// @After:
+// for i := range [10]int{} {
+// 		func() {
+// 			defer f(i)
+//	 	}
+// }
+
+import (
+	"go/ast"
+)
+
+func init() {
+	addChecker(&deferInLoopChecker{}, attrExperimental)
+}
+
+type deferInLoopChecker struct {
+	checkerBase
+}
+
+func (c *deferInLoopChecker) VisitStmt(stmt ast.Stmt) {
+	var body *ast.BlockStmt
+	loop, ok := stmt.(*ast.RangeStmt)
+	if !ok {
+		loop, ok := stmt.(*ast.ForStmt)
+		if !ok {
+			return
+		}
+		body = loop.Body
+	} else {
+		body = loop.Body
+	}
+	for _, s := range body.List {
+		dfr, ok := s.(*ast.DeferStmt)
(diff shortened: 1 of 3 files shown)


lint/deferInLoop_checker.go
```diff
@@ -0,0 +1,51 @@
...
+		}
+		body = loop.Body
+	} else {
+		body = loop.Body
+	}
+	for _, s := range body.List {
+		dfr, ok := s.(*ast.DeferStmt)
+		if ok {
+			c.warn(dfr)
+		}
```
Quasilyte:
```diff
- dfr, ok := s.(*ast.DeferStmt)
- if ok {
- 	c.warn(dfr)
- }
+ if astp.IsDeferStmt(s) {
+ 	c.warn(s)
+ }
```
?

Quasilyte: (This will require changing `cause` type to `ast.Node`)

lint/deferInLoop_checker.go
```diff
@@ -0,0 +1,51 @@
...
+// @Before:
+// for i := range [10]int{} {
+// 		defer f(i) // will be executed only at the end of func
+// }
+//
+// @After:
+// for i := range [10]int{} {
+// 		func() {
+// 			defer f(i)
+//	 	}
```
Quasilyte: Maybe this inner func should accept `i` argument and actually be a call?
Like:
```go
func (i int) { defer f(i) } (i)
```
Otherwise it looks not right.

docs/overview.md
```diff
@@ -549,6 +586,28 @@ func main() {
 ```
 
 
+<a name="lenNil-ref"></a>
+## lenNil
```
Quasilyte: Why is this pops up in multiple PRs?
Can we merge this checker docs (if it's even merged, I don't remember) in a separate PR?
Can be a mistake here, for example, you're pushing your local docs for the checker that was not merged, perhaps?

lint/deferInLoop_checker.go
```diff
@@ -0,0 +1,52 @@
...
+//! Detects defer in loop and warns that it will not be executed till the end of function's scope.
+//
+// @Before:
+// for i := range [10]int{} {
+// 		defer f(i) // will be executed only at the end of func
+// }
+//
+// @After:
+// for i := range [10]int{} {
+// 		func() {
```
Quasilyte: Missing an argument here. `func() => func(i int)`.

PR #369, opened 2018-06-30
makefile: add install option

fexolm:


Quasilyte:
Should probably install `cmd/gocritic` specifically?
Also, `install` is not an `option`, it's a make `target`, IIRC. :)

Quasilyte:
@cristaloleg you mean using relative path?
If gopath is set to the dir containing go-critic, there should be no problem with canonical path.
(But I'm also OK with relative path.)

Issue #345, opened 2018-06-26
lint: add "onlyExported" option for unnamedResult checker


Quasilyte:
One of the main reasons for named results is godoc and API clarity.
For unexported things this may be a little too opinionated for many developers.
We should provide a switch that limits this check to exported functions/methods only.
Don't know what value fits default better. Probably `false`, to force more named results by default.

Quasilyte:
Some signatures from upspin project that trigger this check:
```
func (c *cache) cacheName(de *upspin.DirEntry) (string, string)
func (s *web) accessAll(name upspin.PathName) (bool, bool, error)
func (i *Iterator) GetAndAdvance() (interface{}, interface{}, bool)
```
They could be like this:
```go
func (c *cache) cacheName(de *upspin.DirEntry) (dir string, file string)
func (s *web) accessAll(name upspin.PathName) (readable bool, listable bool, err error)
func (i *Iterator) GetAndAdvance() (key interface{}, value interface{}, valie bool)
```

cristaloleg:
Closed #446

PR #374, opened 2018-07-02
all: config for checkers

### LINKED ISSUE #332: all: ability to provide config for checker
- Easily pass params to linter(possibly via `BindContext` as separate param)
- Must be a JSON file
- Nice to have a command flag to pass JSON

cristaloleg:
Fixes #332 & #345

### DIFF
--- a/cmd/criticize/main.go
+++ b/cmd/criticize/main.go
@@ -1,11 +1,13 @@
 package criticize
 
 import (
+	"encoding/json"
 	"flag"
 	"go/ast"
 	"go/build"
 	"go/parser"
 	"go/types"
+	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
@@ -31,6 +33,7 @@ type linter struct {
 
 	// Command line flags:
 
+	configFile         string
 	withOpinionated    bool
 	withExperimental   bool
 	checkGenerated     bool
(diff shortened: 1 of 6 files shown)


lint/unnamedResult_checker.go
```diff
@@ -20,9 +20,21 @@ func init() {
...
 	checkerBase
+
+	onlyExported bool
+}
+
+func (c *unnamedResultChecker) Init() {
+	v, has := c.ctx.params["onlyExported"]
+	if flag, ok := v.(bool); has && ok {
+		c.onlyExported = flag
+	}
```
Quasilyte: I think this can be simplified to `c.onlyExported = c.ctx.params["onlyExported"]` since default values are the same.

Quasilyte: Hmm.. Maybe this is an idea for checker, too? ;)

Quasilyte: I believe it can be safely simplified to:
```go
c.onlyExported = c.ctx.params["onlyExported"].(bool)
```

cmd/criticize/main.go
```diff
@@ -111,6 +118,25 @@ func parseArgv(l *linter) {
 	}
 }
 
+func (l *linter) LoadConfig() {
+	raw, err := ioutil.ReadFile(l.configFile)
+	if err != nil {
+		return
+	}
```
Quasilyte: This will make specifying non-existing file error impossible to detect.
Maybe we should treat default argument specifically, but if user specified different filename, then exit with error if that config not found?

Quasilyte:
I'm OK with this implementation.
I'll leave some thoughts in #332.

cristaloleg:
Closed in favor of #422

PR #381, opened 2018-07-05
lint/ptrToRefParam add interface

GreenHorse:
interface{} and named interfaces like chan, map, slice are referential
type.

Interface are implemented as struct containing a pointer to the itab
and a pointer to the underlying struct or ptr to struct implementing 
the interface.

This PR is linked to #380 in a sense, but could be view independently.


cristaloleg:
Thanks!

PR #384, opened 2018-07-07
lint: use types.Object in indexOnlyLoop

Quasilyte:
Replaces usages of ast.Object with types.Object.


Quasilyte:
CC @NexoMichael.

PR #396, opened 2018-07-09
lint: add dupSubExpr checker

### LINKED ISSUE #395: lint: add duplicated sub-expr checker
Detects suspicious duplicated sub-expressions.

**Before:**
```go
sort.Slice(xs, func(i, j int) bool {
	return xs[i].v < xs[i].v // Same indexes
})
```
**After:**
```go
sort.Slice(xs, func(i, j int) bool {
	return xs[i].v < xs[j].v // Different indexes
})
```

Quasilyte:
Fixes #395

### DIFF
--- a/lint/dupSubExpr_checker.go
+++ b/lint/dupSubExpr_checker.go
@@ -4,12 +4,12 @@ package lint
 //
 // @Before:
 // sort.Slice(xs, func(i, j int) bool {
-// 	return xs[i].v < xs[i].v // Same index
+// 	return xs[i].v < xs[i].v // Same indexes
 // })
 //
 // @After:
 // sort.Slice(xs, func(i, j int) bool {
-// 	return xs[i].v < xs[j].v // Different index
+// 	return xs[i].v < xs[j].v // Different indexes
 // })
 
 import (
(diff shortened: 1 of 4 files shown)


lint/dupSubExpr_checker.go
```diff
@@ -0,0 +1,121 @@
...
+	if c.isSafe(expr) && c.opSet[expr.Op] && astequal.Expr(expr.X, expr.Y) {
+		c.warn(expr)
+	}
+}
+
+func (c *dupSubExprChecker) resultIsFloat(expr ast.Expr) bool {
+	if typ, ok := c.ctx.typesInfo.TypeOf(expr).(*types.Basic); ok {
+		return typ.Info()&types.IsFloat != 0
+	}
+	return false
```
cristaloleg: Can be:
```
typ, ok := c.ctx.typesInfo.TypeOf(expr).(*types.Basic)
return ok && typ.Info()&types.IsFloat != 0
```

lint/dupSubExpr_checker.go
```diff
@@ -0,0 +1,121 @@
...
+	// Can be extended as needed.
+	//
+	// Note that it is not very strict "safe" as
+	// index expressions are permitted even though they
+	// may cause panics.
+	switch expr := expr.(type) {
+	case *ast.BinaryExpr:
+		return c.isSafe(expr.X) && c.isSafe(expr.Y)
+	case *ast.UnaryExpr:
+		return c.isSafe(expr.X) && expr.Op != token.ARROW
```
cristaloleg: I think it's better to reorder them.
I mean `return expr.Op != token.ARROW && c.isSafe(expr.X)`

lint/dupSubExpr_checker.go
Quasilyte: Thanks!
Not the first time such simplification pops up on the review.
Maybe we can add a checker for it?

lint/dupSubExpr_checker.go
Quasilyte: Makes sense.

Quasilyte:
PTAL.

PR #397, opened 2018-07-09
[WIP] lint: deprecatedUsage checker

cristaloleg:


dominikh:
> AFAIK, staticcheck does this, but I'm not sure if it checks for variables.

You're welcome to ask me ;)

Staticcheck checks any deprecated object that can be specified with a selector expression, i.e. package-level objects (consts, vars, funcs, types), fields and methods. It doesn't detect deprecated objects in the same package as its usage, because that causes false positives. Similarly, functions that are themselves deprecated may use deprecated objects. Another measure to reduce false positives.

PR #405, opened 2018-07-10
lint: rename boolOpt to boolExprSimplify

Quasilyte:
New name avoids confusion related to performance.
This checker is about more readable alternatives,
not more performant ones.

Also updated doc comment to make things clear.


Quasilyte:
> Ok, but maybe name should contains something about expressions? rather than just bool?

Good point. Fixed.

PR #406, opened 2018-07-10
Inlining finding submatches for a warning

nar1k:
Fix for Issue 386


Quasilyte:
Note: we're using golang/go project conventions for commit messages:

So, a better commit message would be:
```
lint: inlined unpackSubmatches, removed function itself

Fixes #386
```

Quasilyte:
And by the way, thanks!

PR #408, opened 2018-07-10
all: separate gometalinter from tests, test checkers in parallel

cristaloleg:


Quasilyte:
Does parallel execution give any boost here?

cristaloleg:
So...we've 2 build stages in this PR: 1 for tests, vet, gocritic and 1 for gometalinter.
And it's depends:
There might be a build when 1st stage is around 1 minute and 2nd is around 2 minutes(https://travis-ci.org/go-critic/go-critic/builds/402287537?utm_source=github_status&utm_medium=notification)

Or both of them are closer to 2 minutes(https://travis-ci.org/go-critic/go-critic/builds/402287528?utm_source=github_status&utm_medium=notification)

Current master is around 3 minutes on average(https://travis-ci.org/go-critic/go-critic/builds/402260731?utm_source=github_status&utm_medium=notification)

I've a proposal to get rid of gometalinter totally. And run it occasionally and fix warnings. WDYT?

PR #411, opened 2018-07-10
lint: defaultCaseOrder

### LINKED ISSUE #412: lint: defaultCaseOrder
Name: `defaultCaseOrder`
Before:
```go
switch a {
case 5:
	// ...
default:  // <- uncool
	// ...
case 42:
	// ...
}
```
After:
```go
switch a {
case 5:
	// ...
case 42:
	// ...
default:
	// ...

}
```

cristaloleg:
Fixes #412


lint/defaultCaseOrder_checker.go
```diff
@@ -0,0 +1,53 @@
...
+	addChecker(&defaultCaseOrderChecker{}, attrSyntaxOnly)
+}
+
+type defaultCaseOrderChecker struct {
+	checkerBase
+}
+
+func (c *defaultCaseOrderChecker) VisitStmt(stmt ast.Stmt) {
+	swtch, ok := stmt.(*ast.SwitchStmt)
+	if !ok || swtch.Body == nil || swtch.Body.List == nil {
```
Quasilyte: I don't think switch statements without body exist...

lint/defaultCaseOrder_checker.go
```diff
@@ -0,0 +1,56 @@
...
+// case x == 10:
+//	// ...
+// default: // <- everything is good
+//	// ...
+// }
+
+import "go/ast"
+
+func init() {
+	addChecker(&defaultCaseOrderChecker{}, attrSyntaxOnly)
```
Quasilyte: Do we want to make it stable right away?
If not, we should probably add `attrExperimental` here.

PR #422, opened 2018-07-11
all: config for checkers

cristaloleg:


cmd/criticize/main.go
```diff
@@ -144,6 +152,33 @@ func parseArgv(l *linter) {
 	}
 }
 
+func (l *linter) LoadConfig() {
+	if _, err := os.Stat(l.configFile); err != nil {
```
Quasilyte: Is `os.IsNotExist` check redundant here?

cmd/criticize/main.go
```diff
@@ -144,6 +152,33 @@ func parseArgv(l *linter) {
...
+		return
+	}
+
+	raw, err := ioutil.ReadFile(l.configFile)
+	if err != nil {
+		log.Fatalf("cannot read config file %s, got error: %s", l.configFile, err)
+		return
```
Quasilyte: Hmm.. Seems like a check candidate.
Return after Fatal-like logging (same for panic logging?).

lint/lint_private.go
```diff
@@ -42,6 +43,38 @@ const (
 	attrVeryOpinionated
 )
 
+type parameters map[string]interface{}
+
+func (p parameters) Int(key string, defaultValue int) int {
```
Quasilyte: Hmm... The disadvantage of such API is that default values are not very apparent and will be spread all across the checker code. But that can work for now! Better than not having config at all.
We can adapt later.

cmd/criticize/main.go
cristaloleg: Hm...it should be warning here ;D

cmd/criticize/main.go
cristaloleg: Maybe..we should just set default config params as an empty string rather than `.gocritic.json`

cmd/criticize/main.go
Quasilyte: Yep.

cmd/criticize/main.go
Quasilyte: Probably. Not critical for the first patch.

PR #425, opened 2018-07-13
lint: teach boolExprSimplify to recognize combined comparisons

Quasilyte:
```
Added rules include:
	x > y || x == y => x >= y
	x < y || x == y => x <= y

Plus their inverted lhs/rhs forms.
```


Quasilyte:
@cristaloleg no. :)
And I'm somewhat glad there are none.

PR #426, opened 2018-07-13
Change 'consider to give name to results'

barbeque:
Looks like a great project! I noticed this warning was a little strangely written, so I decided to stand up this PR to fix it.

Thanks for your time.

### DIFF
--- a/lint/unnamedResult_checker.go
+++ b/lint/unnamedResult_checker.go
@@ -82,5 +82,5 @@ func (c *unnamedResultChecker) typeName(typ types.Type) string {
 }
 
 func (c *unnamedResultChecker) warn(n ast.Node) {
-	c.ctx.Warn(n, "consider to give name to results")
+	c.ctx.Warn(n, "consider giving a name to this result")
 }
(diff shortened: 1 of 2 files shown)


lint/unnamedResult_checker.go
```diff
@@ -82,5 +82,5 @@ func (c *unnamedResultChecker) typeName(typ types.Type) string {
 }
 
 func (c *unnamedResultChecker) warn(n ast.Node) {
-	c.ctx.Warn(n, "consider to give name to results")
+	c.ctx.Warn(n, "consider giving a name to this result")
```
cristaloleg: `results` stands for `resulting parameters`.
Maybe we should leave a multiple form, so: `this results`.
Am I right?

barbeque: `these results` perhaps?

cristaloleg: Oh, yes 😉

cristaloleg:
Will leave it open for a short time, so @Quasilyte will have a look at it. Thank you for your time 🙌

Quasilyte:
Just a note: we're using golang/go conventions for commit messages.
I've squashed with updated message, so it fits these conventions.

PR #439, opened 2018-07-14
 lint: add lostSQLCloseForRows checker

### LINKED ISSUE #430: Check for call Close for sql.Rows data type
This is critical error, if not called Close for sql.Rows data.

But this can happen in different situations:

1. Simple situation:
`rows, err := db.Query(...)
for rows.Next { ... }
rows.Close`

2. Close in defer:
`rows, err := db.Query(...)
if err != nil {...}
defer rows.Close
for rows.Next { ... }`

3. Close in other method:
`rows, err := db.Query(...)
if err != nil {...}
rowsProcessor(rows)

func rowsProcessor(rows *sql.Rows) {
  ...
  rows.Close
}`

I wand do it self, but I not sure about gocritic posibility.

UncleAndy:
lint: add lostSQLCloseForRows checker

Fixes #430

### DIFF
--- a/lint/lostSQLCloseForRows_checker.go
+++ b/lint/lostSQLCloseForRows_checker.go
@@ -1,0 +1,140 @@
+package lint
+
+//! Detects use *sql.Rows without call Close method.
+//
+// @Before:
+// rows, _ := db.Query(...)
+// for rows.Next {
+//   ...
+// }
+//
+// @After:
+// rows, _ := db.Query(...)
+// for rows.Next {
+//   ...
+// }
+// rows.Close
+
+import (
+	"go/ast"
+	"go/types"
+)
+
+func init() {
+	addChecker(&lostSQLCloseForRowsChecker{}, attrExperimental)
+}
+
+type lostSQLCloseForRowsChecker struct {
+	checkerBase
+}
+
+// Warning if sql.Rows local variables (including function parameters):
+// 1. Not using as parameter in other functions call;
+// 2. Not returning in functions results;
+// 3. Not call Close method for variable;
+
+func (c *lostSQLCloseForRowsChecker) VisitFuncDecl(decl *ast.FuncDecl) {
+	// Function parameter variable
+	params := decl.Type.Params
+	var paramVal types.Object
+	for _, p := range params.List {
+		t := c.ctx.typesInfo.TypeOf(p.Type)
+		if t != nil {
+			switch t.String() {
+			case "*database/sql.Rows":
+				paramVal = c.ctx.typesInfo.ObjectOf(identOf(p.Names[0]))
+				break
+			}
+		}
+	}
+
+	localVars := make([]types.Object, 0)
+	returnVars := make([]types.Object, 0)
+	closeVars := make([]types.Object, 0)
+	callVars := make([]types.Object, 0)
+	for _, b := range decl.Body.List {
+		switch b := b.(type) {
+		case *ast.AssignStmt:
+			// Detect local vars with sql.Rows types
+			if b.Lhs != nil {
+				for _, l := range b.Lhs {
+					t := c.ctx.typesInfo.TypeOf(l)
+					if t.String() == "*database/sql.Rows" {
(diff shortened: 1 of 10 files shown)


lint/lostSQLCloseForRows_checker.go
```diff
@@ -0,0 +1,140 @@
...
+// for rows.Next {
+//   ...
+// }
+//
+// @After:
+// rows, _ := db.Query(...)
+// for rows.Next {
+//   ...
+// }
+// rows.Close
```
cristaloleg: `rows.Close()`  😉

lint/lostSQLCloseForRows_checker.go
```diff
@@ -0,0 +1,140 @@
...
+// 3. Not call Close method for variable;
+
+func (c *lostSQLCloseForRowsChecker) VisitFuncDecl(decl *ast.FuncDecl) {
+	// Function parameter variable
+	params := decl.Type.Params
+	var paramVal types.Object
+	for _, p := range params.List {
+		t := c.ctx.typesInfo.TypeOf(p.Type)
+		if t != nil {
+			switch t.String() {
```
cristaloleg: Can be in if above, so `if t != nil && t.String() == "*database/sql.Rows" {`

lint/lostSQLCloseForRows_checker.go
```diff
@@ -0,0 +1,140 @@
...
+		// If local variable present in return or in other functions call or Close present - PASS
+		if !varInList(l, returnVars) &&
+		   !varInList(l, callVars) &&
+		   !varInList(l, closeVars) {
+			c.ctx.Warn(l.Parent(), "local variable db.Rows have not Close call")
+		}
+	}
+}
+
+func varInList(v types.Object, list []types.Object) bool {
```
cristaloleg: `func (c *lostSQLCloseForRowsChecker) varInList`
we're trying to have as less global functions as possible.

lint/testdata/lostSQLCloseForRows/negative_tests.go
```diff
@@ -0,0 +1,144 @@
...
+    db, err := sql.Open("postgres", "")
+    if err != nil {
+        return
+    }
+
+    rows, err := testMethodReturningRows(db)
+    if err != nil {
+        return
+    }
+    defer rows.Close()
```
cristaloleg: Can we add another example, when defer is inside an anonymous function?
```
defer func() {
    rows.Close()
}()
```

lint/testdata/lostSQLCloseForRows/positive_tests.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/// param variable db.Rows have not Close call
+func testPosMethodCloseRows(rows *sql.Rows) {
+	for rows.Next() {
+		var testdata string
+		rows.Scan(&testdata)
+	}
+}
+
```
cristaloleg: Should have `/// param variable db.Rows have not Close call` ?

lint/lostSQLCloseForRows_checker.go
UncleAndy: Ok. I fix it...

lint/lostSQLCloseForRows_checker.go
UncleAndy: Ok...

lint/testdata/lostSQLCloseForRows/negative_tests.go
UncleAndy: Ok. It is possible...

lint/testdata/lostSQLCloseForRows/positive_tests.go
UncleAndy: I move it to negative_tests. It should negative, becouse rows returning.

lint/testdata/lostSQLCloseForRows/negative_tests.go
UncleAndy: I create negative_tests for this case, but temporary comment. For feature release.

PR #447, opened 2018-07-14
lint: add returnAfterLogFatal checker (#444)

### LINKED ISSUE #444: lint: return after fatal/panic
Detects return statements that follow panic/fatal logging.

**Before:**
```go
log.Fatal("exits funciton")
return
```

**After:**
```go
log.Fatal("exits function")
```

breml:
Fixes #444


lint/returnAfterLogFatal_checker.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"go/ast"
+)
+
+func init() {
+	addChecker(&returnAfterLogFatalChecker{}, attrExperimental)
+}
+
+type returnAfterLogFatalChecker struct {
```
Quasilyte: What about calling it `unreachableReturn` or `deadReturn`?
We may add more common functions later, like `os.Exit` to the list, for example.
(Note that I can do renaming after the merge myself, but it's better to decide a name together.)

lint/returnAfterLogFatal_checker.go
```diff
@@ -0,0 +1,46 @@
...
+		case *ast.ExprStmt:
+			if exprStmt, ok := stmt.X.(*ast.CallExpr); ok {
+				switch name := qualifiedName(exprStmt.Fun); name {
+				case "log.Fatal", "log.Fatalf", "log.Fatalln", "log.Panic", "log.Panicf", "log.Panicln":
+					found = name
+				}
+			}
+		case *ast.ReturnStmt:
+			if found != "" {
+				c.warn(stmt, found)
```
Quasilyte: It's probably better to assign empty string to the `found` after `c.warn` call.

PR #456, opened 2018-07-16
lint: fix handling of multi-deref in undered

### LINKED ISSUE #455: lint: underef gives invalid suggestions for **T with underlying pointer
When it suggests to simplify `(**x).y` it proposes `*x.y` instead of `(*x).y`.

Quasilyte:
Fixes #455


Quasilyte:
This change does the following:
If we are about to print suggestion, do not be too hasty to remove parenthesis.
If there is more than one "star" `*`, do not omit parenthesis in suggested form, since Go does only 1-level deref automatically.

PR #462, opened 2018-07-18
lint/emptyFmt: add case for pkg/errors.Wrapf

cristaloleg:
Adding case for this func https://godoc.org/github.com/pkg/errors#Wrapf

### DIFF
--- a/lint/emptyFmt_checker.go
+++ b/lint/emptyFmt_checker.go
@@ -45,8 +45,11 @@ func (c *emptyFmtChecker) VisitExpr(expr ast.Expr) {
 			c.warn(call, "errors.New")
 		}
 	case 2:
-		if name == "fmt.Fprintf" {
+		switch name {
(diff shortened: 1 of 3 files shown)


cristaloleg:
It fails due to unresolved dependency (precisely `pkg/errors`). 
I don't think it's a good idea to add dependency for tests, but this package is heavily used in the community. 
Any ideas?

Quasilyte:
You're not actually checking that the `errors.Wrapf` comes from the `github.com/pkg/errors` package, so it will trigger on arbitrary import with `errors.Wrapf` qualified name. Given that, you maybe you can do a relative import like `import "./errors"` and define a single `Wrapf` function there?

Note that we should think hard about whether this approach of checking only `qualifiedName` suffices form arbitrary package. The proper solution may be to check even more full name like `github.com/pkg/errors.Wrapf`... But in this case, we would not be able to test it without getting the dependency.

PR #466, opened 2018-07-20
lint: make longChain more conservative

### LINKED ISSUE #445: lint: fix longChain false positives
`longChain` checker is broken and needs to be fixed.

Quasilyte:
This is a temporary fix for longChain that had
tons of false positives before this change.

Now it only checks for duplicates in a well-defined places,
like RHS of assignment and switch case clauses.
We can extend this set later, or re-write the whole
algorithm for what matter.

Fixes #445


Quasilyte:
This is work-in-progress version.
Maybe I'll submit a better one later.

lint/internal/lintutil/astmap.go
```diff
@@ -0,0 +1,75 @@
...
+	key ast.Node
+	val interface{}
+}
+
+// AstMap is a simple ast.Node map.
+// Zero value is ready to use set.
+// Can be reused after Clear call.
+// In a simplest case, can be used as AST set.
+type AstMap struct {
+	items []astMapItem
```
cristaloleg: why not a `map[ast.Node]interface{}` ?

PR #467, opened 2018-07-20
fixes #174

### LINKED ISSUE #174: lint: add checker for unnessesary blocks
Warn on blocks without local variables.
**before:**
```go
a := 5
{
     fmt.Print(a)
}
```
**after:**
```go
a := 5
fmt.Print(a)
```

dmirogin:


dmirogin:
Could somebody help me ?
Why test fail here https://travis-ci.org/go-critic/go-critic/jobs/406399595 ? It'is positive, it must be warned, right? What am i doing wrong ?

lint/testdata/unnecessaryBlock/positive_tests.go
```diff
@@ -0,0 +1,8 @@
+package checker_test
+
+func positive () {
+	a := 1
+	{
```
cristaloleg: Here you should have a comment like:
```block doesn't have assignment statements, can be simply deleted```

cristaloleg:
@dmirogin nicely done. Just add a comment and the build will be green 😉

dmirogin:
@cristaloleg Cool. Thanks 👍

Quasilyte:
Note that we're using https://golang.org/doc/contribute.html#commit_messages as a guidance for commit message structure. Squashed commit has updated message to comply to that.

cristaloleg:
@dmirogin thanks for the contribution 😉

PR #468, opened 2018-07-21
lint: suggestions to PR #467

Quasilyte:
- Mark checker as experimental
- Add more positive and negative tests
- Also match DeclStmt to capture const/var/type decls
- Add indentation to docs text (1 missing tab)
- Generate docs


Quasilyte:
CC @dmirogin

PR #478, opened 2018-07-25
lint: don't warn on balanced if-else in elseif checker

### LINKED ISSUE #420: lint: elseif is very noisy (false positives?)
Does report this pattern very frequently:
```go
		if f(r) {
			if wasField {
				spans = append(spans, span{start: fromIndex, end: i})
				wasField = false
			}
		} else {
			if !wasField {
				fromIndex = i
				wasField = true
			}
		}
```
There are lots of similar cases under GOROOT.

Quasilyte:
Change default behavior not to warn on balanced statements.

A balanced if statement is a one where both branches
contain if statement, like in example below:

```go
	if cond {
		if x == 1 {
		}
	} else {
		if x == 2 {
		}
	}
```

gocritic can be instructed to warn on these too by changing
skipBalanced value to false.

Fixes #420

### DIFF
--- a/lint/elseif_checker.go
+++ b/lint/elseif_checker.go
@@ -14,6 +14,8 @@ func init() {
 
 type elseifChecker struct {
 	checkerBase
+
+	skipBalanced bool
 }
 
 func (c *elseifChecker) InitDocumentation(d *Documentation) {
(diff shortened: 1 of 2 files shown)


lint/elseif_checker.go
```diff
@@ -30,15 +32,25 @@ if cond1 {
...
 			return
 		}
-		if astp.IsIfStmt(elseBody.List[0]) {
-			c.warn(stmt.Else)
+		if !astp.IsIfStmt(elseBody.List[0]) {
+			return
+		}
+		balanced := len(stmt.Body.List) == 1 &&
+			astp.IsIfStmt(stmt.Body.List[0])
+		if balanced && c.skipBalanced {
```
cristaloleg: What about keeping everything together?
```
balanced := c.skipBalanced && 
	len(stmt.Body.List) == 1 &&
	astp.IsIfStmt(stmt.Body.List[0])
if balanced {
```

Quasilyte: This is not `balanced` condition if you include everything in it.

Quasilyte: Ah, I see, you suggesting to move `c.skipBalanced` to `balanced` expression.
I don't think it improves readability, actually.

PR #501, opened 2018-07-27
all: initClause implementation

### LINKED ISSUE #492: lint: only assignment is allowed in if init clause
Name: `initClause` 
Before:
```go
if sideEffect(); cond {
}
```
After:
```go
sideEffect()
if cond {
}
```

Checker might be also extended for switches and probably for's

cristaloleg:
Fixes #492

### DIFF
--- a/lint/initClause_checker.go
+++ b/lint/initClause_checker.go
@@ -24,10 +24,9 @@ if cond {
 }
 
 func (c *initClauseChecker) VisitStmt(stmt ast.Stmt) {
-	if expr, ok := c.getInitClause(stmt).(*ast.ExprStmt); ok {
-		if !astp.IsAssignStmt(expr.X) {
-			c.warn(expr, stmt)
-		}
+	initClause := c.getInitClause(stmt)
+	if initClause != nil && !astp.IsAssignStmt(initClause) {
+		c.warn(stmt, initClause)
 	}
 }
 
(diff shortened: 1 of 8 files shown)


lint/initClause_checker.go
```diff
@@ -0,0 +1,48 @@
+package lint
+
+import (
+	"go/ast"
+
+	"github.com/go-toolsmith/astp"
+)
+
+func init() {
+	addChecker(&initClauseChecker{}, attrSyntaxOnly)
```
Quasilyte: Please, add `attrExperimental`.

lint/initClause_checker.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+type initClauseChecker struct {
+	checkerBase
+}
+
+func (c *initClauseChecker) InitDocumentation(d *Documentation) {
+	d.Summary = "Detects non-assignment statements inside if/switch init clause."
+	d.Before = `123`
+	d.After = `123s`
```
Quasilyte: Docs need some work..
I think this patch set is a draft, right?

lint/initClause_checker.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func (c *initClauseChecker) VisitStmt(stmt ast.Stmt) {
+	if expr, ok := c.getInitClause(stmt).(*ast.ExprStmt); ok {
+		if astp.IsCallExpr(expr.X) {
```
Quasilyte: Why not just check if it's not `AssignStmt`?

lint/initClause_checker.go
```diff
@@ -0,0 +1,51 @@
...
+	d.Before = `if sideEffect(); cond {
+}`
+	d.After = `sideEffect()
+if cond {
+}`
+}
+
+func (c *initClauseChecker) VisitStmt(stmt ast.Stmt) {
+	if expr, ok := c.getInitClause(stmt).(*ast.ExprStmt); ok {
+		if !astp.IsAssignStmt(expr.X) {
```
Quasilyte: I'm concerned.
Does this work?

cristaloleg: Sure, why not?

Quasilyte: Actually, you can do this:
```go
if !astp.IsAssignStmt(c.getInitClause(stmt)) {
// warn
}
```

Quasilyte: I'm concerned because `ast.ExprStmt` can't have `ast.AssignStmt` in it's `X` field. Ever.
This is confusing piece of code that looks like unreachable.

PR #503, opened 2018-07-27
all: migrate to golangci-lint

### LINKED ISSUE #500: ci: consider using golangci-lint instead of gometalinter
Can be worthwhile.
It's interesting to see exact CI run time difference with it.
Although I think our pipeline is bottlenecked with installation of packages rather than linting.

Anyway, if first our integration is [golangci-lint](https://github.com/golangci/golangci-lint), we might as well use it inside our project.

cristaloleg:
Fixes #500


Quasilyte:
Thanks for working on this!

cristaloleg:
Waiting for a solution for https://github.com/golangci/golangci-lint/issues/168

.golangci.yml
```diff
@@ -0,0 +1,35 @@
...
+    - misspell
+    - nakedret
+    - staticcheck
+    - structcheck
+    - typecheck
+    - unconvert
+    - unparam
+    - unused
+    - varcheck
+  disable:
```
jirfag: Maybe it’s better to set disable-all:true? It will do the same but will allow easier updates of golangci-lint when new linters will be added

.golangci.yml
```diff
@@ -0,0 +1,35 @@
...
+    - depguard
+    - dupl
+    - gocyclo
+    - govet
+    - interfacer
+    - lll
+    - maligned
+    - megacheck # https://github.com/golangci/golangci-lint/issues/157
+    - prealloc
+  fast: false
```
jirfag: It’s by default false

Makefile
```diff
@@ -16,17 +16,17 @@ docs:
...
 
-ci-gometalinter:
-	go get -u gopkg.in/alecthomas/gometalinter.v2
-	gometalinter.v2 --install
-	gometalinter.v2 --skip=testdata --vendor ./...
+ci-linter:
+	gocritic check-project `pwd`
+	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b ${GOPATH}/bin v1.9.1
+	# golangci-lint run -v
+	golangci-lint run -v --no-config --disable-all --enable=unparam
```
jirfag: Isn’t it better to configure only in config file?

.golangci.yml
cristaloleg: I've decided to mention all the disabled checkers. Just for a clarity :)

Makefile
cristaloleg: Well, atm I'm testing problem with `unparam` (see https://github.com/golangci/golangci-lint/issues/168), that's why I've commented line above.

cristaloleg:
Travis version was downgraded from `1.x` to `1.10.x` due to https://github.com/golang/go/issues/26671

PR #561, opened 2018-08-04
lint: update floatCompare checker

ludweeg:
1. Fix print 'binaryExpr' as second argument
2. Add detection NaNCheckExpr: 'x != x' and InfCheckExpr: 'x + x == x'.
Result: less false-positives.

### DIFF
--- a/lint/floatCompare_checker.go
+++ b/lint/floatCompare_checker.go
@@ -34,8 +34,10 @@ func (c *floatCompareChecker) VisitLocalExpr(expr ast.Expr) {
 		return
 	}
 	if binexpr.Op == token.EQL || binexpr.Op == token.NEQ {
-		if c.isFloatExpr(binexpr) && c.isMultiBinaryExpr(binexpr) &&
-			!c.isNaNCheckExpr(binexpr) && !c.isInfCheckExpr(binexpr) {
+		if c.isFloatExpr(binexpr) &&
+			c.isMultiBinaryExpr(binexpr) &&
+			!c.isNaNCheckExpr(binexpr) &&
+			!c.isInfCheckExpr(binexpr) {
 			c.warn(binexpr)
 		}
 	}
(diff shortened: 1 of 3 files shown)


lint/floatCompare_checker.go
```diff
@@ -30,20 +34,66 @@ func (c *floatCompareChecker) VisitLocalExpr(expr ast.Expr) {
...
+	return astequal.Expr(x, y) && astequal.Expr(y, z)
+}
+
+func (c *floatCompareChecker) isMultiBinaryExpr(binexpr *ast.BinaryExpr) bool {
+	exprx := astutil.Unparen(binexpr.X)
+	expry := astutil.Unparen(binexpr.Y)
+	return astp.IsBinaryExpr(exprx) || astp.IsBinaryExpr(expry)
+}
+
+func identExpr(x ast.Node, y ast.Node) (ast.Expr, *ast.BinaryExpr, bool) {
```
cristaloleg: It will be better to make it as a `floatCompareChecker` method.

lint/floatCompare_checker.go
```diff
@@ -30,20 +34,66 @@ func (c *floatCompareChecker) VisitLocalExpr(expr ast.Expr) {
 		return
 	}
 	if binexpr.Op == token.EQL || binexpr.Op == token.NEQ {
-		typx, ok := c.ctx.typesInfo.TypeOf(binexpr.X).(*types.Basic)
-		if ok && typx.Info()&types.IsFloat != 0 {
+		if c.isFloatExpr(binexpr) && c.isMultiBinaryExpr(binexpr) &&
```
cristaloleg: Stylish: expressions can be splitted by lines 😉

lint/floatCompare_checker.go
cristaloleg: And it might be a `identExpr(x, y ast.Node)` ;D

lint/floatCompare_checker.go
```diff
@@ -30,20 +34,66 @@ func (c *floatCompareChecker) VisitLocalExpr(expr ast.Expr) {
...
+		return x.(ast.Expr), expr2, true
+
+	default:
+		return nil, nil, false
+	}
+}
+
 func (c *floatCompareChecker) warn(expr *ast.BinaryExpr) {
-	var op string
+	op := ">="
```
cristaloleg: So...we do not support `<=`, am I right?

cristaloleg: Ignore this comment :)

Issue #505, opened 2018-07-28
cmd/lintwalk: solve flag forwarding issue between check-package and check-project


quasilyte:
Both sub-commands should have same base set of flags (`-withOpinionated`, etc.).

Flag duplicating starts to become a maintenance burden. Maybe there is a better way.

cristaloleg:
What about creating a `flags` package in `cmd` which will be responsible for flag parsing and validation?

Quasilyte:
That's one of the options.
But lets not be too hasty about it.

cristaloleg:
Sure, just proposing atm.

ewgRa:
Can I take this? For me proposal about shared flags package looks easiest and actually I don't see a problems there, other options probably will be excessive at this moment.

In case if there is will be better solution it will be easy to refactor, package will be well isolated.

btw, what is another options? I see only some cli-specialized library like https://github.com/jessevdk/go-flags. Any other options?

cristaloleg:
"flag" from stdlib, why not?

Quasilyte:
> Can I take this?

Sure.

PR #571, opened 2018-08-14
flag forwarding between commands (#505)

ewgRa:


cmd/criticize/main.go
```diff
@@ -33,24 +34,18 @@ type linter struct {
 
 	// Command line flags:
 
-	configFile string
-
-	withOpinionated    bool
-	withExperimental   bool
-	checkGenerated     bool
-	shorterErrLocation bool
+	fp *flagparser.FlagParser
```
cristaloleg: s/fp/flags ?

lint/flagparser/flag_parser.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func (fp *FlagParser) DisabledCheckers() []string {
+	return strings.Split(fp.Disable, ",")
+}
+
+func (fp *FlagParser) Parse() error {
+	flag.Parse()
+
+	if fp.Enable != EnableAll {
```
cristaloleg: This if can be inverted and everything can be rewritten as a simple swithch 😉

lint/flagparser/flag_parser.go
```diff
@@ -0,0 +1,84 @@
+package flagparser
```
cristaloleg: I think goling is yelling that there is no docs for exported declarations 😉

Quasilyte:
I think this solution can work for now.
I'll check it this evening.
Thanks for submitting this!

lint/flagparser/flag_parser.go
ewgRa: Can you give details?

You mean something like that:
```
	if fp.Enable == EnableAll {
		return nil
	}

	switch {
	case fp.WithExperimental:
		return fmt.Errorf("-withExperimental used with -enable=%q", fp.Enable)
	case fp.WithOpinionated:
		return fmt.Errorf("-withOpinionated used with -enable=%q", fp.Enable)
	}

	return nil
```
? 

I thought about that. It give more room to made a mistake for developer who want to add more checks, he will see this switch, and probably miss that there is "return nil" before and will just expand switch with new case. Such change on my mind made code more fragile against changes.

What do you think?

cmd/criticize/main.go
ewgRa: Done, but why it is better? Is it objective, or subjective?

lint/flagparser/flag_parser.go
ewgRa: Done. Why it is not in travis build?

Also as I don't see there gofmt checks, golint checks, etc.

lint/flagparser/flag_parser.go
```diff
@@ -0,0 +1,84 @@
...
+func (fp *FlagParser) Args() []string {
+	return []string{
+		"-enable", fp.Enable,
+		"-disable", fp.Disable,
+		"-config", fp.ConfigFile,
+		"-withExperimental=" + fmt.Sprint(fp.WithExperimental),
+		"-withOpinionated=" + fmt.Sprint(fp.WithOpinionated),
+		"-failcode", fmt.Sprint(fp.FailureExitCode),
+		"-checkGenerated=" + fmt.Sprint(fp.CheckGenerated),
+		"-shorterErrLocation=" + fmt.Sprint(fp.ShorterErrLocation),
```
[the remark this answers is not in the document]
ewgRa: Done
It works even without "=" or spaces, since this arguments used in exec.Command as argument list. But I add "=" everywhere to made it same everywhere.

Issue #570, opened 2018-08-13
config option work against user experience


ewgRa:
If I run "go run main.go check-package -config a.json fmt" with empty a.json file, I'm expected that no any checkers will be used. There is expectations, that config file also configure checkers, that I want to run.

For now there is unexpected "default" checkers runs.

How about made changes close to user experience and config will be really config?

Quasilyte:
It sounds good to me that there should be a way to configure enabled set via the config.
@cristaloleg, what do you think?

cristaloleg:
Yep, good one.

ewgRa:
closed in #573

PR #573, opened 2018-08-17
config file control enabled checkers (#570)

ewgRa:
PR for #570

There is two commits, one is super simple and made required changes (enabled checkers controlled by config), second one add complexity and restrict to use config option together with "enable", "disable", etc. Some options that can control enabled checkers. 
This on my mind reduce complexity of support and made it more simple to use and understand.

### DIFF
--- a/cmd/criticize/main.go
+++ b/cmd/criticize/main.go
@@ -41,6 +41,51 @@ type linter struct {
 	checkerParams   map[string]map[string]interface{}
 }
 
+// parseEnabledCheckers processes enabled checkers, intersect them with disabled, etc.
+func (l *linter) parseEnabledCheckers() {
+	switch l.flags.Enable {
+	case flagparser.EnableAll:
+		for _, rule := range lint.RuleList() {
+			if rule.Experimental && !l.flags.WithExperimental {
+				continue
+			}
+			if rule.VeryOpinionated && !l.flags.WithOpinionated {
+				continue
+			}
+			l.enabledCheckers = append(l.enabledCheckers, rule.Name())
+		}
+	case "":
+		// Empty slice. Semantically "disable-all".
+		// Can be used to run all pipelines without actual checkers.
+		l.enabledCheckers = []string{}
+	default:
+		// Comma-separated list of names.
+		l.enabledCheckers = l.flags.EnabledCheckers()
+	}
+
+	switch l.flags.Disable {
+	case flagparser.DisableAll:
+		l.enabledCheckers = l.enabledCheckers[:0]
+	case "":
+	// nothing to disable, skip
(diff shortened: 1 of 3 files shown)


cmd/internal/flagparser/flag_parser.go
```diff
@@ -16,28 +18,60 @@ const DisableAll = "all"
...
+	fp.flagSet.BoolVar(&fp.ShorterErrLocation, "shorterErrLocation", true,
+		`whether to replace error location prefix with $GOROOT and $GOPATH`)
+
+	return fp
+}
+
+// FlagParser, where default values different from real ones, used in trick for define that
+// arguments was provided by user, or just was set up by default
+// See https://stackoverflow.com/a/51903637/4143494
+func newDefaultInvertedFlagParser() *FlagParser {
```
cristaloleg: this is a duplicate of `NewFlagParser`, maybe it will be better add a param to `NewFlagParser` which will enable this trick?
(and move comment closer to the trick 😉 )

cmd/criticize/main.go
```diff
@@ -83,7 +85,10 @@ func parseArgv(l *linter) {
 	if len(l.packages) == 0 {
 		blame("no packages specified\n")
 	}
+}
 
+// parseEnabledCheckers processes enabled checkers, intersect them with disabled, etc.
+func parseEnabledCheckers(l *linter) {
```
cristaloleg: `func (l *linter) parseEnabledCheckers() {` ?

cmd/internal/flagparser/flag_parser.go
ewgRa: You mean do something like this?

```
func NewFlagParser(invertedDefaults bool) {
...
    if (invertedDefaults) {
          fp.flagSet.StringVar(&fp.Enable, "enable", "inverted",
		`comma-separated list of enabled checkers`)
    } else {
          fp.flagSet.StringVar(&fp.Enable, "enable", EnableAll,
		`comma-separated list of enabled checkers`)
    }
}
```

Or use the same with ternary workaround variations?

I thought about that before and come to conclusion that there is one disadvantage - mixing this trick logic with normal setup flags, and I don't see advantages of this. And it will really confuse developers. With "if" - will be copypaste again, just in one place, with ternary operator workarounds - will looks heavy.

But I found as I think best solution, invert defaults without hardcode flags, check please, commit https://github.com/go-critic/go-critic/pull/573/commits/9788a5c090a138e95890209e431754cd647fbad3

cmd/criticize/main.go
ewgRa: Good proposal, thanks, done

PR #577, opened 2018-08-22
lint: follow-up for PR576

Quasilyte:
- Rename checker file
- Remove trailing period in docs summary
- Use StmtVisitor
- Reduced nesting in Visit method
- Generate docs


Quasilyte:
CC @breml

breml:
@Quasilyte Thanks for the update. I forgot about `VisitStmt` (aka I copied the wrong checker as the base line). It is now much simpler, I like it. Also thanks for updating the docs. It is really pleasing to have the PRs merged so quickly.

PR #578, opened 2018-08-23
lint: follow-up for PR566

breml:
Add some tests for nested switch-case-fallthrough statements
to prevent the checker logic to be simplified accidentally.
Harmonize tests.

### DIFF
--- a/lint/testdata/emptyFallthrough/negative_tests.go
+++ b/lint/testdata/emptyFallthrough/negative_tests.go
@@ -2,26 +2,48 @@ package checker_tests
 
 import "fmt"
 
-func noWarningsNonEmptyFallthrough(i int) int {
+func noWarningsNonEmptyFallthrough(i int) bool {
 	switch i {
 	case 0:
 		fmt.Print("")
 		fallthrough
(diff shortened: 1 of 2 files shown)


lint/testdata/emptyFallthrough/negative_tests.go
```diff
@@ -2,26 +2,48 @@ package checker_tests
 
 import "fmt"
 
-func noWarningsNonEmptyFallthrough(i int) int {
+func noWarningsNonEmptyFallthrough(i int) bool {
```
cristaloleg: What was the point to change int -> bool?

breml: the point was to harmonize with the negative tests, where always this signature was used. so all tests look similar (it is only a stylistic change).

cristaloleg: Yeah, I thought it's stylistic, just wanted to ensure.
BTW I think we don't have any styleguides for tests 😉

cristaloleg:
Thanks!

Issue #579, opened 2018-08-24
lint: unnecessary rename of import package


breml:
Name: `importPackageName`
Description: Detects when imported package names are unnecessary renamed
Before:

```go
import lint "github.com/go-critic/go-critic/lint"
```

After:

```go
import "github.com/go-critic/go-critic/lint"`
```

PR #580, opened 2018-08-24
lint: add importPackageName checker (#579)

breml:
Fixes: #579

### DIFF
--- a/lint/importPackageName_checker.go
+++ b/lint/importPackageName_checker.go
@@ -1,0 +1,42 @@
+package lint
+
+import (
+	"go/ast"
+	"strings"
+)
+
+func init() {
+	addChecker(&importPackageNameChecker{}, attrExperimental)
+}
+
+type importPackageNameChecker struct {
+	checkerBase
+}
+
+func (c *importPackageNameChecker) InitDocumentation(d *Documentation) {
+	d.Summary = "Detects when imported package names are unnecessary renamed"
+	d.Before = `
+import lint "github.com/go-critic/go-critic/lint"`
+	d.After = `
+import "github.com/go-critic/go-critic/lint"`
+}
+
+func (c *importPackageNameChecker) VisitFile(file *ast.File) {
+	for _, imp := range file.Imports {
+		var pkgName string
+		for _, pkgImport := range c.ctx.pkg.Imports() {
+			if pkgImport.Path() == strings.Trim(imp.Path.Value, `"`) {
+				pkgName = pkgImport.Name()
+				break
+			}
+		}
+
+		if imp.Name != nil && imp.Name.Name == pkgName {
+			c.warn(imp)
+		}
+	}
+}
+
+func (c *importPackageNameChecker) warn(cause ast.Node) {
+	c.ctx.Warn(cause, "unnecessary rename of import package")
+}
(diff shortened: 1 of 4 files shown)


lint/internal/dummy/dummy.go
```diff
@@ -0,0 +1,4 @@
+package dummypkg
```
breml: Not sure, if this package additional package is fine for you.
In order the have a positive check, where the rename of a package actually does rename a package, even if the path of the package does not suggest this, I need such a package. To be able to actually import it, it can not be inside the `testdata` folder.
Therefore I think, that inside the `internal` folder of the `lint` package is the next best place.

lint/testdata/importPackageName/positive_tests.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	/// unnecessary rename of import package
+	fmt "fmt"
+
+	/// unnecessary rename of import package
+	lint "github.com/go-critic/go-critic/lint"
+
+	/// unnecessary rename of import package
+	dummypkg "github.com/go-critic/go-critic/lint/internal/dummy"
```
Quasilyte: Why `dummy` => `dummypkg` rename is unnecessary?

lint/internal/dummy/dummy.go
cristaloleg: Why it cannot be in `testdata` ? Can it be in `lint/testdata/internal` ?

lint/importPackageName_checker.go
```diff
@@ -0,0 +1,41 @@
...
+	d.After = `
+import "github.com/go-critic/go-critic/lint"`
+}
+
+func (c *importPackageNameChecker) VisitFile(file *ast.File) {
+	for _, imp := range file.Imports {
+		var pkgName string
+		for _, pkgImport := range c.ctx.pkg.Imports() {
+			if pkgImport.Path() == strings.Trim(imp.Path.Value, `"`) {
+				pkgName = pkgImport.Name()
```
cristaloleg: Should we break after this line? If not - why? :)

lint/testdata/importPackageName/positive_tests.go
cristaloleg: 'Cause it's name is `dummypkg` already :)

breml: @cristaloleg is right. From the [spec](https://golang.org/ref/spec#Import_declarations):

> If the PackageName is omitted, it defaults to the identifier specified in the package clause of the imported package.

lint/internal/dummy/dummy.go
breml: All files located inside a folder called `testdata` are ignored by the Go compiler. Output of `go help test`:

> The go tool will ignore a directory named "testdata", making it available
to hold ancillary data needed by the tests.

Therefore, it is not possible to import this package in the tests.
In my case (VS code), `goimports` does not find the package (if in folder `testdata`) and therefore removes the import on save.
In order to test (and show case) this special case where `package name` != `directory name` I need this package.

lint/importPackageName_checker.go
breml: You are right, we can leave the loop when the match is found.

Quasilyte:
Sorry for the merge delay.

PR #583, opened 2018-08-26
all: make new command

cristaloleg:


Makefile
```diff
@@ -33,3 +33,6 @@ cover:
 
 install:
 	go install ./cmd/gocritic
+
+new:
+	cd ./cmd/makenew && go run main.go -name=$(filter-out $@,$(MAKECMDGOALS))
```
Quasilyte: What about not doing a `cd` and avoid using paths like `../../` in the makenew?

cmd/makenew/main.go
```diff
@@ -0,0 +1,53 @@
...
+		log.Fatalf("render template: %v", err)
+	}
+
+	filename := checkersPath + *name + "_checker.go"
+
+	if err := ioutil.WriteFile(filename, buf.Bytes(), 0600); err != nil {
+		log.Fatalf("write output file: %v", err)
+	}
+
+	testDir := checkersPath + "testdata/" + *name
```
Quasilyte: What about using `filepath.Join` here?

cmd/makenew/new_checker.go.tmpl
```diff
@@ -0,0 +1,53 @@
+package lint
+
+import (
+	"go/ast"
+	"go/token"
+
+    "github.com/go-toolsmith/astcopy"
+	"github.com/go-toolsmith/astequal"
+	"github.com/go-toolsmith/astfmt"
+	"github.com/go-toolsmith/astp"
```
Quasilyte: Seems like there are issues with import lines alignment.
It's not a big deal, since this is a template, but it looks ugly.

cmd/makenew/main.go
```diff
@@ -0,0 +1,53 @@
...
+	})
+	if err != nil {
+		log.Fatalf("render template: %v", err)
+	}
+
+	filename := checkersPath + *name + "_checker.go"
+
+	if err := ioutil.WriteFile(filename, buf.Bytes(), 0600); err != nil {
+		log.Fatalf("write output file: %v", err)
+	}
```
Quasilyte: You might want to call `format.Source` over the generated code.
Or run `gofmt` over the generated file.

ewgRa:
Is "cmd/makenew/main.go" really needed here?

For me looks a little bit overhead.
It can be done on bash level in Makefile: mkdir -> cp -> sed

PR #586, opened 2018-08-30
lint: add deprecatedComment checker

Quasilyte:
Detects malformed deprecation comments.
When possible, suggests how to fix it.


Quasilyte:
I also plan to include safe checks for typos.
Like in: `Deprcated: ...`, where the format is 100% correct, but there is a typo in `Deprecated` word.

PR #588, opened 2018-08-30
lint: handle simplest typos in deprecatedComment checker

Quasilyte:
There is a big room for improvements, this implementation
is like a stub for further cases.

### DIFF
--- a/lint/deprecatedComment_checker.go
+++ b/lint/deprecatedComment_checker.go
@@ -40,7 +40,6 @@ func (c *deprecatedCommentChecker) Init() {
 	//
 	// TODO(quasilyte): currently it only handles a single missing letter.
 	// Might want to handle other kinds of common misspell/typo kinds.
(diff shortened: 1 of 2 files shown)


Quasilyte:
@cristaloleg, I think it might result in a false positives.
I only remember Levenshtein function out of my head, which will report distance of 1 for `Peprecated`, but this does not look like a typo of `Deprecated`. Not all mismatching letters should result in a "proper diff score". We also don't need a distance, only values we care about are 0 and 1, so it's more like a binary classification, or predicate like `isTypoOf(a, b)` which may report typo probability.

ewgRa:
Maybe look to hunspell integration? Something similar to https://github.com/akhenakh/hunspellgo

Quasilyte:
@ewgRa, well, I think it might be an overkill to spell-check a single word. :sweat_smile:

PR #592, opened 2018-08-30
lint/deprecatedComment: more tests

cristaloleg:


lint/deprecatedComment_checker.go
```diff
@@ -101,8 +103,10 @@ func (c *deprecatedCommentChecker) VisitDocComment(doc *ast.CommentGroup) {
 		}
 
 		// Detect some simple typos.
+		lowerL := strings.ToLower(l)
 		for _, prefixWithTypo := range c.commonTypos {
-			if strings.HasPrefix(l, prefixWithTypo) {
```
Quasilyte: Why not use `upper` var from above and use `ToUpper`?
You can also do ToUpper in `Init` function once, so there is no need to do it per every doc comment.

PR #594, opened 2018-09-01
lint: add inverseMethodCall

### LINKED ISSUE #593: lint/inverseMethodCall:
Name: `inverseMethodCall`
Before:
```go
type foo struct{}
func (f foo) bar() {}

f := foo{}
foo.bar(f)
```
After:
```go
type foo struct{}
func (f foo) bar() {}

f := foo{}
f.bar()
```

Working example: [playground](https://play.golang.org/p/EzhWMaF2HYZ)

Thanks to @matryer 
From: https://youtu.be/5DVV36uqQ4E?t=1334

cristaloleg:
Fixes #593


lint/inverseMethodCall_checker.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	obj := c.ctx.typesInfo.Uses[id]
+	if c.isObjectStruct(obj) {
+		c.warn(call, "XXX")
+	}
+
+}
+
+func (c *inverseMethodCallChecker) isObjectStruct(t types.Object) bool {
```
cristaloleg: There must be a better way to check this :(

lint/inverseMethodCall_checker.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	id, ok := s.X.(*ast.Ident)
+	if !ok {
+		return
+	}
+
+	obj := c.ctx.typesInfo.Uses[id]
+	if c.isObjectStruct(obj) {
+		c.warn(call, "XXX")
```
cristaloleg: Suggestions isn't done yet :(

lint/inverseMethodCall_checker.go
```diff
@@ -0,0 +1,57 @@
...
+
+	s, ok := call.Fun.(*ast.SelectorExpr)
+	if !ok {
+		return
+	}
+
+	id, ok := s.X.(*ast.Ident)
+	if !ok {
+		return
+	}
```
cristaloleg: ^^^ all this can be exported to a method

lint/testdata/inverseMethodCall/positive_tests.go
```diff
@@ -0,0 +1,20 @@
+package checker_test
+
+// https://youtu.be/5DVV36uqQ4E?t=1334
```
cristaloleg: Redundant

lint/inverseMethodCall_checker.go
```diff
@@ -0,0 +1,57 @@
...
+func init() {
+	addChecker(&inverseMethodCallChecker{}, attrExperimental)
+}
+
+type inverseMethodCallChecker struct {
+	checkerBase
+}
+
+func (c *inverseMethodCallChecker) InitDocumentation(d *Documentation) {
+	d.Summary = "Detects inverse method execution"
```
Quasilyte: `Detects method expression call that can be replaced with a method call`

lint/inverseMethodCall_checker.go
```diff
@@ -0,0 +1,57 @@
...
+	d.Before = `type foo struct{}
+f := foo{}
+foo.bar(f)`
+	d.After = `type foo struct{}
+f := foo{}
+f.bar()`
+}
```
Quasilyte: `type foo struct{}` line seems redundant as it's clear that `foo` is a type (literal is used).

lint/inverseMethodCall_checker.go
Quasilyte: We have `identOf`. It can be extended, if you want it.
But it should probably be done in a separate PR.

lint/testdata/_sanity/tests.go
```diff
@@ -200,3 +200,15 @@ func sliceExpressions(xs []int) {
...
 	_ = xs[0:0:0]
 }
+
+type myStruct struct {
+	field string
+}
+
+func (m myStruct) method(a string) {}
+
+func inverseMethodCall() {
```
Quasilyte: `s/inverseMethodCall/methodExprCall` (as in all other places).

PR #603, opened 2018-09-01
Added colored output, json output. And removed unnecessary log prefixes.

AnikHasibul:
Added log message color with `github.com/logrusorgru/aurora` and prettified the log output!


cmd/gocritic/main.go
```diff
@@ -62,7 +62,7 @@ func main() {
 	if cmd == nil {
 		terminate("unknown sub-command: "+sub, printSupportedSubs)
 	}
-	log.SetPrefix(sub + ": ")
+	log.SetPrefix("- ")
```
Quasilyte: Maybe remove prefix completely then?

cmd/criticize/main.go
```diff
@@ -285,7 +286,13 @@ func (l *linter) checkFile(f *ast.File) {
...
 					loc = shortenLocation(loc)
 				}
-				log.Printf("%s: %s: %v\n", loc, c.Rule, warn.Text)
+				log.Printf("%v:\n%v: %v\n",
+					aurora.Red(c.Rule),
+					aurora.Magenta(
+						aurora.Bold(loc),
+					),
+					warn.Text,
+				)
```
Quasilyte: Colored output should be optional and controlled by a flag.
It's also undesirable to break 1-line per warning rule (although some warnings are printing multi-line code, but this is a bug).
It also changes `location: checker: text` to `checker: location: text`. Not sure whether it's better.

cmd/gocritic/main.go
AnikHasibul: Just putting a `-` to make it more understandable that "Here is a new log!"

cmd/criticize/main.go
AnikHasibul:
```txt
checker:
loc/a/tion: text
```

```txt
- ifElseChain:
/home/go/src/logger/text_formatter.go:129:2: should rewrite if-else to switch statement
```

Doesn't this formation look better?

Flag for non-colored output would be great. But also thinking about a `json` flag to work with various editor. As always `vim-go` is the first choice.

cmd/gocritic/main.go
cristaloleg: Anyway we're expecting a log, so `-` doesn't provides useful context.

cmd/criticize/main.go
cristaloleg: I think it should be configurable. Having oneline output simplifies work with a log, not everyone want to see 2 or more lines per warning.
Consider to make it configurable 😉 

> But also thinking about a json flag to work with various editor.

That's a good one, probably. Do you know which other linters provides a json output?

> As always vim-go is the first choice.

Huh, debatable ;D

cmd/criticize/main.go
```diff
@@ -231,14 +233,22 @@ func (l *linter) CheckPackage(pkgPath string) {
 
 	l.ctx.SetPackageInfo(&pkgInfo.Info, pkgInfo.Pkg)
 	for _, f := range pkgInfo.Files {
-		if l.flags.CheckGenerated || !isGenerated(f) {
-			l.ctx.SetFileInfo(l.getFilename(f))
-			l.checkFile(f)
+		if l.flags.IgnoreTests && l.isTestFile(f) {
```
cristaloleg: This is strange. Those changes were already merged into master 🤔

cmd/criticize/main.go
```diff
@@ -285,7 +295,41 @@ func (l *linter) checkFile(f *ast.File) {
...
+					)
+
+				case l.flags.JSONOutput:
+
+					// due to sort capabilities,
+					// using struct instead of map
+					b, err := json.MarshalIndent(struct {
+						Rule     string
+						Location string
+						Warn     string
```
cristaloleg: JSON tags?

cmd/criticize/main.go
```diff
@@ -285,7 +295,41 @@ func (l *linter) checkFile(f *ast.File) {
 				if l.flags.ShorterErrLocation {
 					loc = shortenLocation(loc)
 				}
-				log.Printf("%s: %s: %v\n", loc, c.Rule, warn.Text)
+				switch {
```
cristaloleg: What about extracting this `switch` into a func? WDYT?

cmd/criticize/main.go
```diff
@@ -285,7 +295,41 @@ func (l *linter) checkFile(f *ast.File) {
...
+
+					// due to sort capabilities,
+					// using struct instead of map
+					b, err := json.MarshalIndent(struct {
+						Rule     string
+						Location string
+						Warn     string
+					}{
+						// Using value to ignore c.Rule.Docs and other unnecessary fields
+						Rule:     fmt.Sprintf("%v", c.Rule),
```
cristaloleg: `c.Rule.Name()` will be enough 😉

cmd/gocritic/main.go
```diff
@@ -62,7 +62,7 @@ func main() {
 	if cmd == nil {
 		terminate("unknown sub-command: "+sub, printSupportedSubs)
 	}
-	log.SetPrefix(sub + ": ")
+	log.SetPrefix("")
```
cristaloleg: I think default prefix is an empty string already, isn't it?

cmd/criticize/main.go
```diff
@@ -302,3 +311,41 @@ func shortenLocation(loc string) string {
...
+
+func output(l *linter, rule, loc, warn string) {
+	switch {
+	case l.flags.JSONOutput:
+		// due to sort capabilities,
+		// using struct instead of map
+		b, err := json.MarshalIndent(struct {
+			Rule     string `json:"rule"`
+			Location string `json:"location"`
+			Warning  string `jdon:"warning"`
```
Quasilyte: `s/jdon/json/`

cmd/criticize/main.go
```diff
@@ -302,3 +311,41 @@ func shortenLocation(loc string) string {
...
+			),
+			warn,
+		)
+
+	default:
+		log.Printf("%s:\n%s: %s\n",
+			rule,
+			loc,
+			warn,
+		)
```
Quasilyte: Adding new formats might be okay, but why changing the default one?
There is no "right default format", and I personally don't like a newline inside a warning message.

cmd/gocritic/main.go
Quasilyte: I think the call to `log.SetPrefix` should be removed completely.

Quasilyte:
@AnikHasibul, sorry for long delay. I've been quite busy these days, unfortunately (but this is a no excuse for me).

I want to share my thoughts:
1. It's somewhat distressing that this PR seem to have broken merge with master, as diff shows previously merged code. Could we fix this? If it's unclear how to do it in this PR, maybe pull locally and do a new one? Would be far easier and comfortable.
2. Can we do these 3 things in a separate PRs?
2.1 Removing log prefix. This can be merged immediately. No objections.
2.2 Custom message formatting (or JSON+plain text). I have some ideas, will share with them in form of an issue, where we can discuss it.
2.3 Message colors.

Thanks.

AnikHasibul:
Thanks for the clearance!
Closing this PR. I am opening a new one!

AnikHasibul:
New PR #622

PR #604, opened 2018-09-01
lint: add dupArg checker

### LINKED ISSUE #399: lint: checker that detects calls like copy(dst, dst)
Example:
```go
func withWarning() {
	var dst []int
	var rw readWriter

	/// suspicious call to copy in `copy(dst, dst)`
	copy(dst, dst)
	/// suspicious call to copy in `copy(dst[1:], dst[1:])`
	copy(dst[1:], dst[1:])

	/// suspicious call to append in `append(dst, dst...)`
	dst = append(dst, dst...)

	/// suspicious call to io.Copy in `io.Copy(rw, rw)`
	io.Copy(rw, rw)
}
```

Do we want such kind of checker?
Should it only handle `copy` or more functions, like in example above (How to avoid false positives when then?).

Quasilyte:
Fixes #399

### DIFF
--- a/lint/dupArg_checker.go
+++ b/lint/dupArg_checker.go
@@ -1,0 +1,89 @@
+package lint
+
+import (
+	"go/ast"
+
+	"github.com/go-toolsmith/astequal"
+)
+
+func init() {
+	addChecker(&dupArgChecker{}, attrExperimental)
+}
+
+type dupArgChecker struct {
+	checkerBase
+
+	matchers map[string]func(*ast.CallExpr) bool
+}
+
+func (c *dupArgChecker) InitDocumentation(d *Documentation) {
+	d.Summary = "Detects suspicious duplicated arguments"
+	d.Before = `copy(dst, dst)`
+	d.After = `copy(dst, src)`
+}
+
+func (c *dupArgChecker) Init() {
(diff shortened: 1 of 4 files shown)


lint/dupArg_checker.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func (c *dupArgChecker) InitDocumentation(d *Documentation) {
+	d.Summary = "Detects suspicious duplicated arguments"
+	d.Before = `copy(dst, dst)`
+	d.After = `copy(dst, src)`
+}
+
+func (c *dupArgChecker) Init() {
+	newMatcherFunc := func(pairs [][2]int) func(*ast.CallExpr) bool {
```
cristaloleg: As I understood `pairs` contains a pair of argument indexes, right?
Can we make name more descriptive? or add a comment for a `newMatcherFunc`?

Quasilyte: How about `indexes` and adding a comment for `newMatcherFunc`?

cristaloleg: Good enough :)

Quasilyte: `argIndexPairs`? Give a hint. :)
Different algorithm is a hint too.

PR #606, opened 2018-09-01
lint: don't warn on *slice in ptrToRefParam 

Quasilyte:


Quasilyte:
Damn. Seems like I've branched not from the master. :sweat_smile:

PR #613, opened 2018-09-03
lint: use precomputed imports map in importShadow

### LINKED ISSUE #532: lint: handle renamed imports more efficiently
Looking at `importShadow`, it's clear that the current way of handling renamed imports is rather inefficient and somewhat clumsy. We need something better that does not require defs traversal per definition (close to n^2).

Quasilyte:
Fixes #532

### DIFF
--- a/lint/lint.go
+++ b/lint/lint.go
@@ -93,7 +93,7 @@ type Checker struct {
 func (c *Checker) Check(f *ast.File) []Warning {
 	c.ctx.warnings = c.ctx.warnings[:0]
 	if c.ctx.require.pkgObjects {
-		resolvePkgObjects(&c.ctx, f)
+		resolvePkgObjects(c.ctx.Context, f)
 	}
 	c.walker.WalkFile(f)
 	return c.ctx.warnings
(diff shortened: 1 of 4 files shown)


lint/lint_private.go
```diff
@@ -193,3 +194,17 @@ func addChecker(c abstractChecker, attrs ...checkerAttribute) {
 	}
 	checkerPrototypes[rule.name] = proto
 }
+
+func resolvePkgObjects(ctx *context, f *ast.File) {
+	ctx.pkgObjects = make(map[*types.PkgName]string)
```
cristaloleg: `make(map[*types.PkgName]string, len(f.Imports))` ?
Just to make it explicit what we're expecting, unless you're against this.

lint/lint.go
```diff
@@ -123,6 +126,19 @@ type Context struct {
...
 	// Arch-dependent.
 	sizesInfo types.Sizes
+
+	// require records what optional resources are required
+	// by the checkers set that use this context.
+	//
+	// Every require fields makes associated context field
+	// to be properly initialized.
+	// For example, Context.require.pkgObjects => Context.pkgObjects.
+	require struct {
```
cristaloleg: Damn, that's a good pattern!

lint/lint_private.go
Quasilyte: OK.

Quasilyte:
We definitely don't want to  disable `unused` from CI pipeline, 
so I think we can afford some workarounds in this case.

PR #621, opened 2018-09-06
all: clean go.sum

cristaloleg:


Quasilyte:
I'm not opposed to that, but I thought that `go.sum` is append-only log.
https://research.swtch.com/vgo-repro

> The go.modverify file is a log of the hash of all versions ever encountered: lines are only added, never removed. If we update rsc.io/sampler to v1.3.1, then the log will now contain hashes for both versions

cristaloleg:
From `go help modules`:

```
The go command maintains, in the main module's root directory alongside
go.mod, a file named go.sum containing the expected cryptographic checksums
of the content of specific module versions. Each time a dependency is
used, its checksum is added to go.sum if missing or else required to match
the existing entry in go.sum.

The go command maintains a cache of downloaded packages and computes
and records the cryptographic checksum of each package at download time.
In normal operation, the go command checks these pre-computed checksums
against the main module's go.sum file, instead of recomputing them on
each command invocation. The 'go mod verify' command checks that
the cached copies of module downloads still match both their recorded
checksums and the entries in go.sum.
```

Quasilyte:
I doesn't say that you should/must clean up (tidy) the log. :)

PR #622, opened 2018-09-06
Added colored output with -color flag (default true), json output wit…

AnikHasibul:
Still has the new line in colored output. And used pretty json formation `json.MarshalIndent` . 
No new line in default output.

### DIFF
--- a/cmd/criticize/main.go
+++ b/cmd/criticize/main.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/go-critic/go-critic/cmd/internal/flagparser"
 	"github.com/go-critic/go-critic/lint"
+	"github.com/logrusorgru/aurora"
 	"golang.org/x/tools/go/loader"
 )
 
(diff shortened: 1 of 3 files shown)


AnikHasibul:
Since the logs are getting printed separately, so the current json formation is invalid:

```txt
{
    "rule": "captLocal",
    "location": "/home/go/src/boom/example.go:14:7",
    "warning": "`Whoops' should not be capitalized"
}
{
    "rule": "captLocal",
    "location": "/home/go/src/boom/example2.go:33:26",
    "warning": "`Oops' should not be capitalized"
}
```

It should be like this:

```txt
[{
    "rule": "captLocal",
    "location": "/home/go/src/boom/example.go:14:7",
    "warning": "`Whoops' should not be capitalized"
},
{
    "rule": "captLocal",
    "location": "/home/go/src/boom/example2.go:33:26",
    "warning": "`Oops' should not be capitalized"
}]
```

Notice the `[`,`]` and the `,` between each log.

We can add the `,` between each log by the std `json.MarshalIndent` function. But the `[]` is seems hard to implement.

Any way to achieve this?

I am thinking about to buffering the whole output when the `-json` flag is present.

Then marshal the whole buffer, and print as json output. WDYT?

Quasilyte:
It's easier when there is example concrete use case, so it's possible to evaluate the impact of one solution other another. Could you share where you're planning to use it?

PR #636, opened 2018-09-08
#635 Rename unusedParam to blankParam

### LINKED ISSUE #635: lint: rename unusedParam to blankParam
If we're keeping this checker, it should be renamed in order to avoid confusion.
"unusedParam" sounds like something that `unused` linter from staticcheck would do.
"blankParam" is as precise, but avoids that confusion.

See also #516

Atul9:
closes #635 
Rename unusedParam to blankParam


Quasilyte:
I think you need to rename `lint/testdata` folder too.

Atul9:
@Quasilyte @cristaloleg done. I have updated the PR.

PR #648, opened 2018-09-15
lint: add unlabelStmt checker

### LINKED ISSUE #214: lint: add loopUnlabel checker
Detects labeled loops that can be re-written as non-labeled loop.

**Before:**
```go
Outer:
	for _, table := range tables {
		for _, x := range table {
			if x == 0 { 
				continue Outer // Check next table
			}
			nonZero++
		}
	}
```
**After:**
```go
	for _, table := range tables {
		for _, x := range table {
			if x == 0 { 
				break // Check next table
			}
			nonZero++
		}
	}
```
There can be more patterns that can be trivially re-written, but they don't come to my mind right now.
I don't like labeled statements and `continue X`/`break X`, especially if they can be avoided.

AFAIK, no linters that do such check exist.

Quasilyte:
Fixes #214

### DIFF
--- a/lint/unlabelStmt_checker.go
+++ b/lint/unlabelStmt_checker.go
@@ -1,0 +1,162 @@
+package lint
+
+import (
+	"go/ast"
+	"go/token"
+)
+
+func init() {
+	addChecker(&unlabelStmtChecker{}, attrExperimental)
+}
+
+type unlabelStmtChecker struct {
+	checkerBase
+}
+
+func (c *unlabelStmtChecker) InitDocumentation(d *Documentation) {
+	d.Summary = "Detects redundant statement labels"
+	d.Before = `
+derp:
+for x := range xs {
+	if x == 0 {
+		break derp
+	}
+}`
+	d.After = `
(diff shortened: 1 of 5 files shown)


Quasilyte:
Seems like docs were not re-generated for a while now..

lint/unlabelStmt_checker.go
```diff
@@ -0,0 +1,162 @@
...
+	case *ast.RangeStmt, *ast.ForStmt, *ast.SwitchStmt, *ast.TypeSwitchStmt, *ast.SelectStmt:
+		return true
+	default:
+		return false
+	}
+}
+
+// blockStmtOf returns body of specified node.
+//
+// TODO(quasilyte): handle other statements and see if it can be useful
```
cristaloleg: Really useful

Quasilyte: It can be used in `singleCaseSwitchChecker`.
Switch statement can be removed in favor or `blockStmtOf` call, I think.

PR #650, opened 2018-09-15
makefile: fix "go vet" invocation cmd, ignore testdata

Quasilyte:


Quasilyte:
Should fix recent CI build failure.

Issue #459, opened 2018-07-17
all: performanceAttribute


cristaloleg:
We might be interested into grouping checkers under this attribute, it will allows users to see what checkers can suggest performance improvements.
Also we might add cmd flag for those checkers also.

Quasilyte:
I had such idea before but @fexolm convinced me that we don't need them.

cristaloleg:
Any arguments?

Quasilyte:
I've CCed him. I don't remember myself.

Quasilyte:
Actually, now I think that performance-related checks should be disabled by default and should require `withPerformance` flag to be activated.

Quasilyte:
@cristaloleg, what do you think?

cristaloleg:
👍

cristaloleg:
Resolved in #652

PR #652, opened 2018-09-17
lint: introduce performance attribute

Quasilyte:
Updates #459

We still need to reflect this attr in docs (overview.md) and add relevant flags for gocritic cmd.

### DIFF
--- a/lint/appendAssign_checker.go
+++ b/lint/appendAssign_checker.go
@@ -11,7 +11,7 @@ import (
 )
 
 func init() {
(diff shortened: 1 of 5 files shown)


Quasilyte:
`makedocs` does not handle this attr yet.
I think we can do it separately.
This is my small step for that. :)
I also don't know how to mark them. Maybe additional emoji near checker name?

cristaloleg:
What about 🚀 ?

Quasilyte:
@cristaloleg, I don't know. :)
Maybe text tag like `[perf]` would work better here.

PR #660, opened 2018-09-24
lint/internal/astwalk: use walker name instead of visitor

Quasilyte:
Visitor and walker are different things in that package.
Some types that should have been called walkers are
called visitors. It's time to fix that.


Quasilyte:
Ping.

Issue #684, opened 2018-10-20
lint: add negative tests for typeSwitchVar


Quasilyte:
There are only [positive tests](https://github.com/go-critic/go-critic/blob/master/lint/testdata/typeSwitchVar/positive_tests.go) for the `typeSwitchVar` checker.
There should be at least some negative tests that verify that it doesn't trigger on the simplified type switches.

GaryPWhite:
I'm on it!

agneum:
I've added some negative tests. Please check the PR: https://github.com/go-critic/go-critic/pull/688

PR #688, opened 2018-10-21
lint: add negative tests to typeSwitchVar

agneum:
I've added some negative tests to typeSwitchVar.
The issue: https://github.com/go-critic/go-critic/issues/684

### DIFF
--- a/lint/testdata/typeSwitchVar/negative_tests.go
+++ b/lint/testdata/typeSwitchVar/negative_tests.go
@@ -1,0 +1,93 @@
+package checker_test
+
+type negativeTypeSwitchVarPoint struct {
+	x int
+	y int
+}
+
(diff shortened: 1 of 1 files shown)


lint/testdata/typeSwitchVar/negative_tests.go
```diff
@@ -0,0 +1,93 @@
...
+			return int(x)
+		}
+	}
+	return 0
+}
+
+func negativeTypeSwitchVarF5(x, y, z interface{}) int {
+	switch x := x.(type) {
+	case int:
+		switch y:= y.(type) {
```
Quasilyte: Looks like this file is not gofmt'ed, but anyway, LGTM.
I can run a formatter by myself.
Thanks you for the contribution!

PR #692, opened 2018-10-23
lint: add structLitKeyOrder checker

### LINKED ISSUE #691: lint: check consistent fields order
```go
type Foo struct {
    bar string
    baz int
}

f := Foo{
    baz: 42,
    bar: "answer"
}
```

Linter can suggest using the same order as in type declaration.

Quasilyte:
Fixes #691

### DIFF
--- a/lint/structLitKeyOrder_checker.go
+++ b/lint/structLitKeyOrder_checker.go
@@ -1,0 +1,75 @@
+package lint
+
+import (
+	"go/ast"
+	"go/types"
+
+	"github.com/go-critic/go-critic/lint/internal/lintutil"
+	"github.com/go-toolsmith/astp"
+)
+
+func init() {
+	addChecker(&structLitKeyOrderChecker{}, attrVeryOpinionated, attrExperimental)
+}
+
+type structLitKeyOrderChecker struct {
+	checkerBase
+}
+
+func (c *structLitKeyOrderChecker) InitDocumentation(d *Documentation) {
+	d.Summary = "Detects struct literal keys order that does not match declaration order"
+	d.Before = `
+type foo struct{ x, y int }
+v := foo{y: y, x: x}`
+	d.After = `
+type foo struct{ x, y int }
+v := foo{x: x, y: y}`
+}
+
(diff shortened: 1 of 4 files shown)


lint/testdata/structLitKeyOrder/positive_tests.go
```diff
@@ -0,0 +1,27 @@
+package checker_test
+
+type point struct {
+	x int
+	y int
+	z int
```
cristaloleg: What about complex type?
```
type Foo struct {
	x Bar
}

type Bar struct {
	y string
}
```

cristaloleg: And what about embedded structs?

lint/structLitKeyOrder_checker.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	// Check until we find a field that is misplaced.
+	// Misplaced field is a field that comes after a field
+	// that had higher declaration order value.
+	maxOrder := -1
+	for _, elt := range lit.Elts {
+		kv := elt.(*ast.KeyValueExpr)
+		order := fieldsOrder[lintutil.AsIdent(kv.Key).Name]
+		if order > maxOrder {
```
cristaloleg: I think this `if` can be inverted easily 😉

Quasilyte: I don't get it, unfortunately.
Can you show me the code?
For me it seems like inverting here does not provide any improvement.

lint/testdata/structLitKeyOrder/positive_tests.go
Quasilyte: Done.

lint/structLitKeyOrder_checker.go
cristaloleg: Yeah, sorry for that. I mean fast exit:
```
if order <= maxOrder {
	c.warn(expr)
	return
}
maxOrder = order
```
but it's about style mostly 😉

Issue #698, opened 2018-10-28
all: remove redundant newlines before `}`


Quasilyte:
I found at least one such place:
https://github.com/go-critic/go-critic/blob/e56208e371282d8583805003e4f7606552292569/lint/unlambda_checker.go#L57-L59

There can be more.
Such empty newlines should be removed.

Yatheesan:
removed new lines in all_checkers_test.go
https://github.com/go-critic/go-critic/pull/700

IPedgE:
searched all files and removed them #701

PR #701, opened 2018-10-28
* removed all redundant newlines before `}` Issue: #698

IPedgE:
searched all files with grep - regex: ^\s*$\s*}


Quasilyte:
Also, congrats with completing hacktoberfest challenge. :)

Issue #703, opened 2018-10-28
lint: replace lint/internal/lintutil coerce.go with astcast package


Quasilyte:
Replace usages of `lintutil.AsXxx` with `astcast.ToXxx`.
See https://github.com/go-toolsmith/astcast.

Quasilyte:
`IsIdent` replaced in #708.

PR #710, opened 2018-10-30
lint: replace lint/internal/lintutil coerce.go with astcast package

weeellz:
Updates #703.

### DIFF
--- a/lint/appendCombine_checker.go
+++ b/lint/appendCombine_checker.go
@@ -4,7 +4,8 @@ import (
 	"go/ast"
 	"go/token"
 
-	"github.com/go-critic/go-critic/lint/internal/lintutil"
+	"github.com/go-toolsmith/astcast"
+
 	"github.com/go-toolsmith/astequal"
 )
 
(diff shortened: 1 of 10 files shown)


weeellz:
I think I'll replace all functions in this PR.

Quasilyte:
@weeellz, okay. :)

Don't forget to remove functions that are no longer used.

weeellz:
@Quasilyte, what to do with `AsUnaryExprOp` which `AsUnaryExprOp is like AsUnaryExpr, but also checks for op token.` There is no replacement function in astcast that checks for op token.

Quasilyte:
> @Quasilyte, what to do with AsUnaryExprOp which AsUnaryExprOp is like AsUnaryExpr, but also checks for op token.?
There is no replacement function in astcast that checks for op token.

Only functions from `corece.go` can be easily replaced.
The function you described can be remained as is.

> And there is IsNil function in coerce.go. What should I do with it?

Can be done in a separate PR if you want to.
Basically:
```go
// before
x := lintutil.AsIdent(a)
return lintutil.IsNil(x)

// after
x := astcast.AsIdent(a)
return x == astcast.NilIdent
```

Quasilyte:
> But AsUnaryExprOp is in coerce.goconfused.

OK, then leave them for now.

PR #713, opened 2018-10-30
lint: remove redundant empty lines

bubblesupreme:


lint/testdata/_sanity/tests.go
```diff
@@ -28,12 +28,6 @@ func empty2() {
 
 func external()
 
-const ()
-
-var ()
-
-type ()
-
```
Quasilyte: Files under `lint/testdata` are not "real" code, they used to test the linter, so it's better to keep them as they were, especially sanity tests. Their aim is to feed potentially unexpected (but valid) code to the linter.

Quasilyte:
Hmm, I pressed "resolve" by mistake.

Quasilyte:
To make things simpler, I'll merge it as is and revert testdata contents separately.
Thanks for the contribution.

Issue #493, opened 2018-07-27
lint: make boolExprSimplify checker stable


Quasilyte:
See #424 for stabilization process overview.
Checker to be stabilized: https://go-critic.github.io/overview#boolExprSimplify-ref
This issue marks checker with higher priority for stabilization.

Quasilyte:
Blocked by #673.

PR #741, opened 2018-12-29
checkers: avoid eager copy in boolExprSimplify

Quasilyte:
Don't copy AST expressions those type is not bool.
We still do a lot of copying, but at least it's only
performed for boolean-typed expressions.

Refs #493

### DIFF
--- a/checkers/boolExprSimplify_checker.go
+++ b/checkers/boolExprSimplify_checker.go
@@ -36,8 +36,12 @@ type boolExprSimplifyChecker struct {
 }
 
 func (c *boolExprSimplifyChecker) VisitExpr(x ast.Expr) {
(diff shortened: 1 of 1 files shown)


checkers/boolExprSimplify_checker.go
```diff
@@ -36,8 +36,12 @@ type boolExprSimplifyChecker struct {
 }
 
 func (c *boolExprSimplifyChecker) VisitExpr(x ast.Expr) {
-	// TODO: avoid eager copy?
-	// Can't be stable until wasted copying is fixed.
+	// Throw away non-bool expressions and awoid redundant
```
Quasilyte: Damn, a typo is there: `s/awoid/avoid/`.

PR #747, opened 2019-01-01
checkers: make docStub more reliable

### LINKED ISSUE #640: lint: false positives in docStub
docStub prints warnings for signatures like:
```go
// Issue 24791.
func TestClonePipe(t *testing.T) {
}
```
It shouldn't.

Quasilyte:
Fixes #640

### DIFF
--- a/checkers/docStub_checker.go
+++ b/checkers/docStub_checker.go
@@ -2,7 +2,9 @@ package checkers
 
 import (
 	"go/ast"
+	"go/token"
 	"regexp"
+	"strings"
 
 	"github.com/go-lintpack/lintpack"
 	"github.com/go-lintpack/lintpack/astwalk"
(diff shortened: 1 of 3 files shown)


checkers/docStub_checker.go
```diff
@@ -25,25 +27,69 @@ func Foo() {}
...
+		}
+	}
+}
+
+func (c *docStubChecker) visitDoc(decl ast.Node, sym *ast.Ident, doc *ast.CommentGroup, article bool) {
+	if !sym.IsExported() || doc == nil {
+		return
+	}
+	line := strings.TrimSpace(doc.List[0].Text[len("//"):])
+	if article {
```
cristaloleg: Can we enable it by default?
`The Foo does bla-bla` seems possible.

PR #754, opened 2019-01-08
all: use modules for coverage

cristaloleg:


Quasilyte:
I think it can be closed.

Issue #767, opened 2019-01-14
all: fix coverage regression


Quasilyte:
It's probably related to incorrect cover profile collection.
We need to adjust Makefile/travis file to make sure code under `/testdata` doesn't count as uncovered code.

Quasilyte:
Interesting.
https://coveralls.io/builds/21077856/source?filename=checkers%2FunnecessaryBlock_checker.go

Only init code is marked as covered, like all of our end2end tests are not included into that coverage profile, because if they would, at the very least checker constructor should execute.

Quasilyte:
When I run `make cover` locally, I get these results:
![image](https://user-images.githubusercontent.com/6286655/51177986-5af95800-18d2-11e9-9315-06c12313e786.png)

Quasilyte:
The coverage decreased in abd843651d8be92bce8680e13867458eb68993dd.
It's probably some modulues<->CI interaction that make things go wrong.
I can't reproduce low coverage on a local machine, even if I use modules and gocovmerge.

PR #769, opened 2019-01-15
WIP: make: collect coverprofile separately from goveralls

Quasilyte:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ ci:
 	@if [ "$(TEST_SUITE)" = "linter" ]; then make ci-linter; else make ci-tests; fi
 
 ci-tests:
-	go test -v -race -count=1 ./...
+	go test -v -race -count=1 -coverprofile=coverage.out ./...
 
 ci-linter:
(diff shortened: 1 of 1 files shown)


Quasilyte:
This might be a solution for #767

Makefile
```diff
@@ -1,5 +1,7 @@
 .PHONY: test test-checker ci cover tools docs
 
+export GO111MODULE := on
```
cristaloleg: This might be a problem for contributors, but it's easy to fix.
So no problem.

go.mod
```diff
@@ -9,5 +9,6 @@ require (
 	github.com/go-toolsmith/astp v0.0.0-20180903215135-0af7e3c24f30
 	github.com/go-toolsmith/strparse v0.0.0-20180903215201-830b6daa1241
 	github.com/go-toolsmith/typep v0.0.0-20181030061450-d63dc7650676
+	github.com/mattn/goveralls v0.0.2 // indirect
```
cristaloleg: I'm not a big fun of an another dependency (which is a tool for testing only).
But for now it's ok.

PR #772, opened 2019-01-18
add hexLiterals checker

bontequero:


checkers/testdata/hexLiterals/negative_tests.go
```diff
@@ -0,0 +1 @@
+package checker_test
```
cristaloleg: What about negative tests like:
```
0xFF
0xff
"0xfF"
```
?

checkers/hexLiterals_checker.go
```diff
@@ -0,0 +1,73 @@
...
+	ctx *lintpack.CheckerContext
+}
+
+func (c *hexLiteralChecker) VisitExpr(x ast.Expr) {
+	if !astp.IsBasicLit(x) {
+		return
+	}
+	v := astcast.ToBasicLit(x)
+
+	if !strings.Contains(v.Value, "0X") && !strings.Contains(v.Value, "0x") {
```
cristaloleg: I think we need `strings.HasPrefix` here. Am I missing something?

checkers/hexLiterals_checker.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"go/ast"
+	"strings"
+	"unicode"
+
+	"github.com/go-toolsmith/astcast"
+
+	"github.com/go-toolsmith/astp"
+
```
cristaloleg: Remove empty lines, please 😉

Issue #427, opened 2018-07-13
lint: add stringXbytes checker that detects redundant conversions between these two


Quasilyte:
There are few special cases for string<->bytes conversions that can simplify the code.
For example, one may copy `string` to `[]byte` without conversion.

----

Detects redundant string<->[]byte conversions.

**Before:**
```go
v := []byte{'a', 'b', 'c'} // Only simple runes
copy(b, []byte(s))
```

**After:**
```go
v := []byte("abc")
copy(b, s)
```

----

Though we may also want to match this:
```go
v := string('x')
// =>
v := "x"
```
And that would require a different check name.
Or maybe we can do a separate checker like `stringXrune` that may cover even more string<->rune cases (can't give other examples right away).

Quasilyte:
This one is not very complicated and can be a good one for first-time contributors.
Let's mark it appropriately and leave unassigned for a while.

Quasilyte:
I've verified that `gometalinter --enable-all` does not detect these.

PR #773, opened 2019-01-18
Implementing detection redundant string<->[]byte conversions

7phs:
Implementing detection redundant string<->[]byte conversions for `copy(b, []byte(s))` case #427

### DIFF
--- a/checkers/stringXbytes_checker.go
+++ b/checkers/stringXbytes_checker.go
@@ -1,0 +1,50 @@
+package checkers
+
+import (
+	"go/ast"
+	"strings"
+
+	"github.com/go-lintpack/lintpack"
+	"github.com/go-lintpack/lintpack/astwalk"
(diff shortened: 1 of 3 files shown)


checkers/stringXbytes_checker.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+type stringXbytes struct {
+	astwalk.WalkHandler
+	ctx *lintpack.CheckerContext
+}
+
+func (c *stringXbytes) VisitExpr(expr ast.Expr) {
+	if x, ok := expr.(*ast.CallExpr); ok {
+		switch name := qualifiedName(x.Fun); name {
```
Quasilyte: Could you rewrite 1-case switch to an if statement?
Then the CI will pass and we can merge it.

7phs: fixed

PR #774, opened 2019-01-18
checkers: followup for #773

Quasilyte:
- Use "style" tag instead of "diagnostic"
- Use byteCast.Args[0] instead of strings woodoo


Quasilyte:
CC @7phs

PR #787, opened 2019-02-01
ci: enable integration tests

### LINKED ISSUE #785: ci: install lintpack and enable integration tests on travis
Right now integration tests are executed only locally.
We need to enable them on CI.

Quasilyte:
Fixes #785


Quasilyte:
I think 1 test file is missing so integration tests should fail.

Quasilyte:
Now we actually run all tests on CI instead of their subset. :sweat_smile:

PR #806, opened 2019-02-17
checkers: add newDeref checker

Quasilyte:
Finds *new(T) expressions and suggests to simplify them
by using T zero value instead.

For example:

	*new(string) => ""
	*new(bool)   => false
	*new(myInt)  => myInt(0)

### DIFF
--- a/checkers/internal/lintutil/zero_value.go
+++ b/checkers/internal/lintutil/zero_value.go
@@ -1,0 +1,93 @@
+package lintutil
+
+import (
+	"go/ast"
+	"go/constant"
+	"go/token"
+	"go/types"
+)
+
+// IsZeroValue reports whether x represents zero value of its type.
+//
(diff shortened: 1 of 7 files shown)


checkers/internal/lintutil/zero_value.go
```diff
@@ -0,0 +1,93 @@
...
+		return &ast.CallExpr{
+			Fun:  typeExpr,
+			Args: []ast.Expr{&ast.Ident{Name: "nil"}},
+		}
+
+	case *types.Array, *types.Struct:
+		return &ast.CompositeLit{Type: typeExpr}
+	}
+
+	return nil
```
cristaloleg: `default` ? :)

Quasilyte: You can send a pr that fixes that. 😉

Issue #779, opened 2019-01-23
caseOrder: error: runtime error: invalid memory address or nil pointer dereference


saromanov:
Hi! I found, that in the case if type on the switch statement doesn't exist, go-critic returns panic
### Example
```go
type test struct {
	value interface{}
}

func do(t *test) {
	switch v := t.value.(type) {
	case MyType: //MyType is not defined
		fmt.Println("there is not exist")
	case string:
		fmt.Println("value is string")
	default:
		fmt.Println()
	}
}
```
### Output
```sh
gocritic check-project -withExperimental .
caseOrder: error: runtime error: invalid memory address or nil pointer dereference
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x790878]

goroutine 95 [running]:
github.com/go-critic/go-critic/cmd/criticize.(*linter).checkFile.func1.1(0xc0003a2130, 0xc000381a40)
	/home/test/gowork/src/github.com/go-critic/go-critic/cmd/criticize/main.go:311 +0x138
panic(0x853d20, 0xc4cb20)
	/usr/local/go/src/runtime/panic.go:513 +0x1b9
github.com/go-critic/go-critic/lint.(*caseOrderChecker).checkTypeSwitch(0xc000142290, 0xc00012dbc0)
	/home/test/gowork/src/github.com/go-critic/go-critic/lint/caseOrder_checker.go:59 +0x138
github.com/go-critic/go-critic/lint.(*caseOrderChecker).VisitStmt(0xc000142290, 0x954a40, 0xc00012dbc0)
	/home/test/gowork/src/github.com/go-critic/go-critic/lint/caseOrder_checker.go:37 +0x76
github.com/go-critic/go-critic/lint/internal/astwalk.(*stmtWalker).WalkFile.func1(0x9516e0, 0xc00012dbc0, 0x950a20)
	/home/test/gowork/src/github.com/go-critic/go-critic/lint/internal/astwalk/stmt_walker.go:19 +0x90
go/ast.inspector.Visit(0xc000035cd0, 0x9516e0, 0xc00012dbc0, 0xc0001c9540, 0xc0001c9540)
	/usr/local/go/src/go/ast/walk.go:373 +0x3a
go/ast.Walk(0x94f620, 0xc000035cd0, 0x9516e0, 0xc00012dbc0)
[...]

PR #808, opened 2019-02-25
caseOrder: error: runtime error: invalid memory address or nil pointer dereference

weeellz:
Update #779.


Quasilyte:
We can merge that but I wonder why type checking passes for that code.
I would expect linter not to be executed on such packages that have type errors.
Needs investigation.

Issue #810, opened 2019-02-27
Initial go get fails


DryHumour:
An initial attempt to get go-critic will fail with a git submodule error.  This failure does, however, leave the git clone checked out and a subsequent get will work (presumably because an empty go-critic sub-sub-sub-directory was created by git submodule during the initial clone).
```
$ go get -v github.com/go-critic/go-critic/...
github.com/go-critic/go-critic (download)
# cd /home/nschelle/go/src/github.com/go-critic/go-critic; git submodule update --init --recursive
fatal: No url found for submodule path 'go-critic' in .gitmodules
package github.com/go-critic/go-critic/...: exit status 128
```
My git-fu is weak, but I think the problem is that commit fc92fdb14a33766600962e970854229d0aace500 (or b9373569160643640c0cfa345c5b6c8d549ad36d?) introduced some sort of phantom submodule without a .gitmodules.
```
$ git ls-files --stage | grep 160000
160000 b9373569160643640c0cfa345c5b6c8d549ad36d 0	go-critic
```

Helcaraxan:
This can now be closed I think, @Quasilyte. :slightly_smiling_face:

Issue #812, opened 2019-02-27
make CI green again


Quasilyte:
Build on master started to fail.

https://travis-ci.org/go-critic/go-critic/jobs/499426168

```
github.com/golang/lint@v0.0.0-20190227174305-5b3e6a55c961: parsing go.mod: unexpected module path "golang.org/x/lint"
```

Quasilyte:
Checking out last successful build (b9373569160643640c0cfa345c5b6c8d549ad36d) doesn't help. Seems like some dependency has changed causing the build to fail.

Helcaraxan:
I have not yet found which dependency but the root cause is golang/lint#397.

Some dependency somewhere must have a deprecated `go.mod` that is referencing `github.com/golang/lint` instead of `golang.org/x/lint`.

Quasilyte:
We could track that dependency and, for example, vendor it, and keep it until upstream is fixed (or it can be better to use specific revision, before it started to cause problems).

@cristaloleg, what is your opinion?

Quasilyte:
https://github.com/go-critic/go-critic/pull/813#issuecomment-468219688

>  @Quasilyte have actually just left a comment there about the root cause. Have not been able to trace the offending dependency yet (no appropriate tooling for that exists yet AFAIK with go modules).

Seems like an idea for a new tool.
I think we should at least ping @myitcv or, better yet, post it on `golang-tools`.
Maybe other people faced the same problem.

myitcv:
Why are you doing `go get -u` here?

https://github.com/go-critic/go-critic/blob/fc92fdb14a33766600962e970854229d0aace500/Makefile#L24

You already have `github.com/go-lintpack/lintpack` as a module requirement, no?

https://github.com/go-critic/go-critic/blob/fc92fdb14a33766600962e970854229d0aace500/go.mod#L4

The `-u` flag here is likely the cause:

> The -u flag instructs get to update dependencies to use newer minor or
patch releases when available. Continuing the previous example,
'go get -u A' will use the latest A with B v1.3.1 (not B v1.2.3).

(from `go help get` in Go 1.12)

Helcaraxan:
@Quasilyte have added the fix to #813.

Helcaraxan:
@myitcv good point indeed. Am new to this project so had not had a look at the script. Thank you very much for the help!

myitcv:
You're welcome!

PR #813, opened 2019-02-28
Remove erroneously added submodule

Helcaraxan:
As it says on the tin. Fix for #810.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,11 +26,11 @@ ci-tests:
 ci-linter:
 	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b ${GOPATH}/bin v1.15.0
 	golangci-lint run -v
-	go get github.com/Quasilyte/go-consistent
+	go install github.com/Quasilyte/go-consistent
 	go-consistent ./...
 
 cover:
-	go get -u github.com/mattn/goveralls
+	go install github.com/mattn/goveralls
 	goveralls -package github.com/go-critic/go-critic/checkers -coverprofile=coverage.out -service travis-ci -repotoken ${COVERALLS_TOKEN}
 
 install:
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/go-critic/go-critic
 
 require (
+	github.com/Quasilyte/go-consistent v0.0.0-20181230194409-8f8379e70f99
 	github.com/go-lintpack/lintpack v0.5.2
 	github.com/go-toolsmith/astcast v0.0.0-20181028201508-b7a89ed70af1
 	github.com/go-toolsmith/astcopy v0.0.0-20180903214859-79b422d080c4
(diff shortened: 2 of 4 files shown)


Quasilyte:
CI fails due to #812, which is a top1 concern right now. :)
I'll try to fix that today, so we can continue on all the PRs.
If you have the idea why that issue exists, your help can be invaluable.

Helcaraxan:
@Quasilyte have actually just left a comment there about the root cause. Have not been able to trace the offending dependency yet (no appropriate tooling for that exists yet AFAIK with go modules).

Helcaraxan:
Now also contains the fix for #810.

Makefile
```diff
@@ -21,17 +21,17 @@ ci:
 	@if [ "$(TEST_SUITE)" = "linter" ]; then make ci-linter; else make ci-tests; fi
 
 ci-tests:
-	go get -u github.com/go-lintpack/lintpack/...
+	go get github.com/go-lintpack/lintpack/...
```
myitcv: I'm not sure you need the `go get` at all here. You're working in module mode. If the dependency is not available it will be fetched.

Makefile
```diff
@@ -21,17 +21,17 @@ ci:
...
 	go test -v -race -count=1 -coverprofile=coverage.out ./...
 
 ci-linter:
 	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b ${GOPATH}/bin v1.12.3
 	golangci-lint run -v
-	go get -u github.com/Quasilyte/go-consistent
+	go get github.com/Quasilyte/go-consistent
```
myitcv: You could add `github.com/Quasilyte/go-consistent` as a [tool dependency](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module); then simply `go install github.com/Quasilyte/go-consistent`

Makefile
```diff
@@ -21,17 +21,17 @@ ci:
...
 	go-consistent ./...
 
 cover:
-	go get -u github.com/mattn/goveralls
+	go get github.com/mattn/goveralls
```
myitcv: Same point about a tool dependency for `github.com/mattn/goveralls`

Helcaraxan:
All very good points @myitcv. Thank you. Am still learning to get to grips with modules as my workplace is not yet using them.

Quasilyte:
Hmm. I already merged #815.
Could you resolve conflicts, please?
Thank you. :)

myitcv:
@Helcaraxan 👍 

> Am still learning to get to grips with modules as my workplace is not yet using them.

The modules wiki (https://github.com/golang/go/wiki/Modules) is an excellent place to start, as is the `cmd/go` docs (https://golang.org/cmd/go/#hdr-Modules__module_versions__and_more).

Helcaraxan:
Here we go. With the added simplifications to the Makefile as a bonus.

myitcv:
Nice work @Helcaraxan!

In a follow up PR you might want to add a `go mod tidy` followed by a check that git remains "clean". Taking https://github.com/myitcv/gobin/blob/b3a85f068656d9366007505254233cead223beb2/.travis.yml#L50-L63 as an example:

```
go mod tidy

# If you are testing Go 1.11: https://github.com/golang/go/issues/27868#issuecomment-431413621
go list all > /dev/null

test -z "$(git status --porcelain)" || (git status; git diff; false)
```

Helcaraxan:
@myitcv done indeed in #817.

PR #816, opened 2019-02-28
tools.go: add a note about "dependency tracking"

Quasilyte:
This will help us to remember why this file is there
and how this pattern is called.


Quasilyte:
CC @Helcaraxan

PR #817, opened 2019-02-28
Add CI check to keep 'go.mod' clean

Helcaraxan:
As per https://github.com/go-critic/go-critic/pull/813#issuecomment-468242697.


Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: test test-checker ci cover tools docs
+.PHONY: test test-checker test-goroot docs ci ci-tidy ci-tests ci-linter cover install
```
Quasilyte: Thanks for the extra cleanups. :blush:

PR #814, opened 2019-02-28
New dupImports checker

Helcaraxan:
Am proposing a new "style" checker that verifies that you are not importing the same package twice in the same Go file under different aliases.

This is not using one of the `astwalk.WalkerFor*` interfaces (which are proxies for `lintpack.FileWalker`) as those do not give access to a file's imports. Instead it directly reimplements the `lintpack.FileWalker` interface itself.

Have added the required positive and negative test files as well.

Example output:
```text
 -> gocritic check -enable=dupImports checkers/testdata/dupImports/positive_tests.go
./checkers/testdata/dupImports/positive_tests.go:4:8: dupImports: package is imported 3 times under different aliases on lines 4, 8 and 10
./checkers/testdata/dupImports/positive_tests.go:8:2: dupImports: package is imported 3 times under different aliases on lines 4, 8 and 10
./checkers/testdata/dupImports/positive_tests.go:10:2: dupImports: package is imported 3 times under different aliases on lines 4, 8 and 10
```

PR #818, opened 2019-03-01
checkers: follow-up for #814

quasilyte:
- Using ast file (fPos) is not needed.
  It's possible to get Pos line by using a node and FileSet only.

- Trailing period in the Summary is redundant.
  We need to write tests for doc objects validation.

- Using plural forms in checker names is uncommon.
  s/dupImports/dupImport/

- Before/After examples should be as minimal, as possible.
  They are not required to be a complete Go program.
  Simplified them a bit.

- Code that emits a warning and builds up a message
  should be moved to "warn" function.
  This is a convention.


quasilyte:
CC @Helcaraxan

PR #857, opened 2019-05-23
fix dependencies

quasilyte:
Refs https://github.com/quasilyte/go-consistent/issues/34


go.mod
```diff
@@ -1,16 +1,17 @@
 module github.com/go-critic/go-critic
 
+go 1.13
```
cristaloleg:
```suggestion
go 1.12
```

quasilyte: I'll use go 1.12 to re-generate those files. :)

PR #864, opened 2019-07-17
build for master as well

Fale:
add to the CI master so that it's always possible to check that go-critic is ready for the next go

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: go
 go:
   - 1.x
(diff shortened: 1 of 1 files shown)


cristaloleg:
That's interesting https://travis-ci.org/go-critic/go-critic/jobs/560111962#L512

cristaloleg:
Thanks again @Fale

Fale:
Thank you :)

PR #866, opened 2019-08-22
add regexpPattern checker

quasilyte:
Currently only checks whether the pattern contains
an unescaped "." near known top level domain.

So, `google.com` is probably flawed and
`google\.com` is considered to be correct.

It's not a security issue most of the time, but
it's better to keep that surface as thin as possible
to avoid potential problems.

### DIFF
--- a/checkers/regexpPattern_checker.go
+++ b/checkers/regexpPattern_checker.go
@@ -64,6 +64,5 @@ func (c *regexpPatternChecker) VisitExpr(x ast.Expr) {
 }
 
 func (c *regexpPatternChecker) warnDomain(cause ast.Expr, domain string) {
-	c.ctx.Warn(cause, "'.%s' should probably be '\\.%s'",
-		domain, domain)
+	c.ctx.Warn(cause, "'.%s' should probably be '\\.%s'", domain, domain)
 }
--- a/checkers/regexpPattern_checker.go
+++ b/checkers/regexpPattern_checker.go
@@ -1,0 +1,68 @@
+package checkers
+
+import (
+	"go/ast"
+	"go/constant"
+	"regexp"
+	"strings"
+
+	"github.com/go-lintpack/lintpack"
+	"github.com/go-lintpack/lintpack/astwalk"
+)
+
+func init() {
+	var info lintpack.CheckerInfo
+	info.Name = "regexpPattern"
+	info.Tags = []string{"diagnostic", "experimental"}
+	info.Summary = "Detects suspicious regexp patterns"
+	info.Before = "regexp.MustCompile(`google.com|yandex.ru`)"
+	info.After = "regexp.MustCompile(`google\\.com|yandex\\.ru`)"
+
+	collection.AddChecker(&info, func(ctx *lintpack.CheckerContext) lintpack.FileWalker {
+		domains := []string{
+			"com",
+			"org",
+			"info",
+			"net",
+			"ru",
+			"de",
+		}
+
+		domainRE := regexp.MustCompile(`[^\\]\.(` + strings.Join(domains, "|") + `)\b`)
+		return astwalk.WalkerForExpr(&regexpPatternChecker{
+			ctx:      ctx,
+			domainRE: domainRE,
+		})
+	})
+}
+
+type regexpPatternChecker struct {
+	astwalk.WalkHandler
+	ctx *lintpack.CheckerContext
+
+	domainRE *regexp.Regexp
+}
+
+func (c *regexpPatternChecker) VisitExpr(x ast.Expr) {
+	call, ok := x.(*ast.CallExpr)
+	if !ok {
+		return
+	}
+
+	switch name := qualifiedName(call.Fun); name {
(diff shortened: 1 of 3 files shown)


checkers/regexpPattern_checker.go
```diff
@@ -0,0 +1,68 @@
...
+func init() {
+	var info lintpack.CheckerInfo
+	info.Name = "regexpPattern"
+	info.Tags = []string{"diagnostic", "experimental"}
+	info.Summary = "Detects suspicious regexp patterns"
+	info.Before = "regexp.MustCompile(`google.com|yandex.ru`)"
+	info.After = "regexp.MustCompile(`google\\.com|yandex\\.ru`)"
+
+	collection.AddChecker(&info, func(ctx *lintpack.CheckerContext) lintpack.FileWalker {
+		domains := []string{
```
cristaloleg: Those domains might be at the end of the file.
BTW, do we need to extend them via checker param?

checkers/regexpPattern_checker.go
```diff
@@ -0,0 +1,68 @@
...
+			"com",
+			"org",
+			"info",
+			"net",
+			"ru",
+			"de",
+		}
+
+		domainRE := regexp.MustCompile(`[^\\]\.(` + strings.Join(domains, "|") + `)\b`)
```
cristaloleg:
```suggestion
		allDomains := strings.Join(domains, "|")
		domainRE := regexp.MustCompile(`[^\\]\.(` + allDomains + `)\b`)
```

checkers/regexpPattern_checker.go
```diff
@@ -0,0 +1,68 @@
...
+	domainRE *regexp.Regexp
+}
+
+func (c *regexpPatternChecker) VisitExpr(x ast.Expr) {
+	call, ok := x.(*ast.CallExpr)
+	if !ok {
+		return
+	}
+
+	switch name := qualifiedName(call.Fun); name {
```
cristaloleg:
```suggestion
	switch qualifiedName(call.Fun) {
```
we don't use name and the func is self-explanatory enough 😉

checkers/regexpPattern_checker.go
```diff
@@ -0,0 +1,68 @@
...
+		s := constant.StringVal(cv)
+		if m := c.domainRE.FindStringSubmatch(s); m != nil {
+			c.warnDomain(call.Args[0], m[1])
+		}
+	}
+}
+
+func (c *regexpPatternChecker) warnDomain(cause ast.Expr, domain string) {
+	c.ctx.Warn(cause, "'.%s' should probably be '\\.%s'",
+		domain, domain)
```
cristaloleg: Hm, do we need this line break?

checkers/regexpPattern_checker.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/go-lintpack/lintpack"
+	"github.com/go-lintpack/lintpack/astwalk"
+)
+
+func init() {
+	var info lintpack.CheckerInfo
+	info.Name = "regexpPattern"
+	info.Tags = []string{"diagnostic", "experimental"}
+	info.Summary = "Detects suspicious regexp patterns"
+	info.Before = "regexp.MustCompile(`google.com|yandex.ru`)"
```
cristaloleg: Should we look for `http`/`https`/`ftp` prefixes?

checkers/testdata/regexpPattern/negative_tests.go
```diff
@@ -0,0 +1,11 @@
+package checker_test
+
+import (
+	"regexp"
+)
+
+func domainDots() {
+	regexp.MustCompile(`google\.com`)
+
+	regexp.CompilePOSIX(`yandex\.ru|radio.yandex\.ru`)
```
cristaloleg:
```suggestion
	regexp.CompilePOSIX(`yandex\.ru|radio.yandex\.ru`)
	
	regexp. MustCompile(`ftp://apple\.net|https:/golang\.org`)
```

checkers/testdata/regexpPattern/positive_tests.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"regexp"
+)
+
+func domainDots() {
+	/*! '.com' should probably be '\.com' */
+	regexp.MustCompile(`google.com`)
+
+	/*! '.ru' should probably be '\.ru' */
+	regexp.CompilePOSIX(`yandex.ru|radio.yandex.ru`)
```
cristaloleg:
```suggestion
	regexp.CompilePOSIX(`yandex.ru|radio.yandex.ru`)
	
	regexp. MustCompile(`ftp://apple.net|https://golang.org`)
```

checkers/regexpPattern_checker.go
quasilyte: I think it's safe and more useful not to restrict regexps to specific prefixes.

checkers/regexpPattern_checker.go
quasilyte: It's a good idea to add custom domains support, but I think this list is good enough for the beginning.

PR #867, opened 2019-08-22
checkers: handle ^ and $ anchors

quasilyte:
- Report unescaped use of ^ not in the beginning of the pattern
- Report unescaped use of $ not in the end of the pattern

Both warnings are disabled if pattern has multi-line mode
enabled (achieved by flags parsing).

### DIFF
--- a/checkers/regexpPattern_checker.go
+++ b/checkers/regexpPattern_checker.go
@@ -31,8 +31,10 @@ func init() {
 		allDomains := strings.Join(domains, "|")
 		domainRE := regexp.MustCompile(`[^\\]\.(` + allDomains + `)\b`)
 		return astwalk.WalkerForExpr(&regexpPatternChecker{
-			ctx:      ctx,
-			domainRE: domainRE,
+			ctx:           ctx,
+			domainRE:      domainRE,
+			beginAnchorRE: regexp.MustCompile(`[^\\^]\^`),
+			endAnchorRE:   regexp.MustCompile(`[^\\]\$.+`),
 		})
 	})
 }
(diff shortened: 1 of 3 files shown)


checkers/regexpPattern_checker.go
```diff
@@ -41,7 +43,16 @@ type regexpPatternChecker struct {
...
+	domainRE      *regexp.Regexp
+	beginAnchorRE *regexp.Regexp
+	endAnchorRE   *regexp.Regexp
+}
+
+type regexpFlags struct {
+	i bool // case-insensitive
+	m bool // multi-line mode: ^ and $ match begin/end line in addition to begin/end text
+	s bool // let . match \n
+	U bool // ungreedy: swap meaning of x* and x*?, x+ and x+?, etc
```
cristaloleg: is it ok to make it exported but leave other unexported?

quasilyte: I wanted to reflect flag byte in the field name. The struct is not exported, so I don't see a problem. :)
From regexp docs:
```
i              case-insensitive (default false)
m              multi-line mode: ^ and $ match begin/end line in addition to begin/end text (default false)
s              let . match \n (default false)
U              ungreedy: swap meaning of x* and x*?, x+ and x+?, etc (default false)
```

cristaloleg: I mean: do we need this small (but important) inconsistency? ;D

quasilyte:
This was a good attempt, but `$` and `^` have their uses in alternation groups. Proper handling would require deeper analysis, maybe `syntax.Parse` can be used to do a better job here.

PR #871, opened 2019-09-03
checkers: fix panic on shadowing "copy" builtin

### LINKED ISSUE #870: checkers/dupArg: panic on shadowing "copy"
Running "gocritic check" on a package like this panics:
```go
package foo

func Foo(x map[string]interface{}) map[string]interface{} {
	cp := copy(x)
	return cp
}

func copy(a map[string]interface{}) map[string]interface{} {
	return a
}
```

The panic appears to be in the `dupArg` checker, which assumes `copy` has at least two params.

reillywatson:
The dupArg checker assumes "copy" has at least two arguments. Builtins can be redefined though, so this assumption may not be true!

Fixes #870.

### DIFF
--- a/checkers/dupArg_checker.go
+++ b/checkers/dupArg_checker.go
@@ -24,6 +24,9 @@ func init() {
 		// args[xIndex] and args[yIndex] are equal.
 		newMatcherFunc := func(xIndex, yIndex int) func(*ast.CallExpr) bool {
 			return func(call *ast.CallExpr) bool {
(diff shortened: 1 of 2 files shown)


quasilyte:
Thanks for fixing this. There are several places where we consider stdlib/builtins as something reliable, the `builtinShadow` checker attempts to discourage users from shadowing builtins, although it only works for variables now and does not trigger on a function named `copy`, for example.

PR #886, opened 2019-11-17
Contrib polish

snargleplax:


snargleplax:
"How to add a new checker" instructions were a bit out-of-date, so I updated them and did some other polish to the text.

PR #887, opened 2019-11-18
Add whyNoLint checker.

snargleplax:
This checker inspects machine-readable //nolint comments (per the
conventions from https://github.com/golangci/golangci-lint#nolint) to
ensure that they are accompanied by human-readable explanatory
comments on the same line.

The intention is to enable teams to enforce a policy of not disabling
linters without providing an explicit justification.

### DIFF
--- a/checkers/whyNoLint_checker.go
+++ b/checkers/whyNoLint_checker.go
@@ -1,0 +1,52 @@
+package checkers
+
+import (
+	"go/ast"
+	"regexp"
+	"strings"
+
+	"github.com/go-lintpack/lintpack"
+	"github.com/go-lintpack/lintpack/astwalk"
+)
+
+func init() {
+	info := lintpack.CheckerInfo{
+		Name:    "whyNoLint",
+		Tags:    []string{"style", "experimental"},
+		Summary: "Ensures that //nolint comments include an explanation",
+		Before:  `//nolint`,
+		After:   `//nolint // reason`,
+	}
+	re := regexp.MustCompile(`^// *nolint(?::[^ ]+)? *(.*)$`)
+
+	collection.AddChecker(&info, func(ctx *lintpack.CheckerContext) lintpack.FileWalker {
+		return astwalk.WalkerForComment(&whyNoLintChecker{
+			ctx: ctx,
+			re:  re,
+		})
+	})
+}
+
+type whyNoLintChecker struct {
+	astwalk.WalkHandler
+
+	ctx *lintpack.CheckerContext
+	re  *regexp.Regexp
+}
+
+func (c whyNoLintChecker) VisitComment(cg *ast.CommentGroup) {
+	if strings.HasPrefix(cg.List[0].Text, "/*") {
+		return
+	}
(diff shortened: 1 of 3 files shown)


checkers/whyNoLint_checker.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/go-lintpack/lintpack"
+	"github.com/go-lintpack/lintpack/astwalk"
+)
+
+func init() {
+	info := lintpack.CheckerInfo{
+		Name:    "whyNoLint",
+		Tags:    []string{"style", "experimental"},
+		Summary: "Ensures that //nolint comments include an explanation",
```
cristaloleg:
```suggestion
		Summary: "Ensures that `//nolint` comments include an explanation",
```
wdyt?

checkers/whyNoLint_checker.go
```diff
@@ -0,0 +1,52 @@
...
+	if strings.HasPrefix(cg.List[0].Text, "/*") {
+		return
+	}
+	for _, comment := range cg.List {
+		sl := c.re.FindStringSubmatch(comment.Text)
+		if len(sl) < 2 {
+			continue
+		}
+
+		if !strings.HasPrefix(sl[1], "//") || len(strings.TrimPrefix(sl[1], "//")) == 0 {
```
cristaloleg: Minor, but `sl[1]` might be extracted to the variable 😉

cristaloleg:
BTW, as I understand, `//nolint` directive is supported only by golanci-lint (and probably archived gometalinter)(maybe staticheck or so too), isn't it?

checkers/whyNoLint_checker.go
snargleplax: Sure. I hadn't realized this was meant to be markdown syntax, but it looks like there's plenty of precedent for this in other checkers. Should we add something to the GoDoc for `lintpack.CheckerInfo.Summary` to make that more clear for checker implementers? Would the expectation be for general-case markdown syntax, or just the backticks for fixed-width (all I've found examples of)?

Anyway, will add for this.

snargleplax:
Looks like staticcheck has their [own convention](https://staticcheck.io/docs/#ignoring-problems):

```
//lint:ignore Check1[,Check2,...,CheckN] reason
```

Since it appears that accommodates a reason string as well, the approach here could be straightforwardly extended to cover their convention.

I don't have experience with any other such drivers, so I don't know what else may be out there to support. Most likely it'd be something else simple to support as in the case above.

PTAL.

quasilyte:
We can add staticcheck-style comments in a separate PR. :)

checkers/whyNoLint_checker.go
quasilyte: It's not markdown-only, as `gocritic doc <checker>` gives that info in a terminal as plain text.

cristaloleg:
Thanks you @snargleplax and sorry for a delay.

snargleplax:
No problem at all, thanks everyone.

PR #891, opened 2019-11-30
all: use Github Actions

cristaloleg:


cristaloleg:
Might be unfinished, but I would like to merge it @quasilyte

PR #908, opened 2020-03-19
makefile: correct linter.version argument

sosiska:
Same as https://github.com/go-critic/go-critic/pull/890


quasilyte:
I updated binaries on a release page.

PR #920, opened 2020-04-14
checkers: improve badRegexp checker

quasilyte:
- Improved flag flow checking.
  Now it can handle flag clearing.

- Added more tests (mostly to negative_tests.go)

- Added checks for dangling/redundant `^` and `$`.

- Fix false positive with altAnchors.


checkers/badRegexp_checker.go
```diff
@@ -66,11 +68,64 @@ func (c *badRegexpChecker) checkPattern(pat string) {
...
+	// so we start from the empty flag set.
+	c.flagStates = append(c.flagStates, regexpFlagState{})
 
+	c.markGoodCarets(re.Expr)
+	c.markGoodDollars(re.Expr)
 	c.walk(re.Expr)
 }
 
+func (c *badRegexpChecker) markGoodCarets(e syntax.Expr) {
+	canSkip := func(e syntax.Expr) bool {
```
cristaloleg, line 84: can be a method, isn't it?

quasilyte: I failed to find a good name for it since it's only used inside this method context.. :(

cristaloleg: ok, lgtm

Issue #926, opened 2020-05-18
False positive for unnecessaryDefer


bombsimon:
It seems like the check for `unnecessaryDefer` only checks if the defer statement is the last statement in a block, no matter what block it is. Because of this it's not possible to set conditional defer statements.

Noticed this when a build in [`golangci-lint`](https://github.com/golangci/golangci-lint/pull/1148) failed. Here's a smaller example to reproduce.

```go
package main

import "fmt"

func main() {
    if true {
        defer func() {
            fmt.Println("something was true")
        }()
    }

    fmt.Println("more code before return")
}
```

```sh
$ gocritic check -enable='unnecessaryDefer' ./...
./main.go:7:3: unnecessaryDefer: defer func(){...}(...) is placed just before return
```

quasilyte:
Fixed in #927

PR #927, opened 2020-05-18
Fix false positives for unnecessaryDefer. #926

bombsimon:
By keeping track of the block type we can differentiate between
functions and other blocks. This way we can enure that only blocks
ending with explicit return or end of functions yelds warnings.


quasilyte:
Since it passes the tests and fixes the referenced issue, maybe we can merge it right away.

I can make a better review if I'll actually look at the entire checker code which I can't do right now. :)

quasilyte:
If I'll have anything to add, there will be a follow-up.
@bombsimon feel free to submit more changes and/or tests. :)

PR #929, opened 2020-05-26
checkers: add sortSlice checker

quasilyte:
Find suspicious sort slice calls.

Sorting `slice` argument with a callback that uses
another slice is usually incorrect, as only one
slice will be modified by Slice call inplace.

### DIFF
--- a/checkers/sortSlice_checker.go
+++ b/checkers/sortSlice_checker.go
@@ -1,0 +1,135 @@
+package checkers
+
+import (
+	"go/ast"
+	"go/token"
+
+	"github.com/go-critic/go-critic/checkers/internal/lintutil"
+	"github.com/go-lintpack/lintpack"
+	"github.com/go-lintpack/lintpack/astwalk"
+	"github.com/go-toolsmith/astcast"
+	"github.com/go-toolsmith/astequal"
+	"github.com/go-toolsmith/typep"
+	"golang.org/x/tools/go/ast/astutil"
+)
+
+func init() {
+	var info lintpack.CheckerInfo
+	info.Name = "sortSlice"
+	info.Tags = []string{"diagnostic", "experimental"}
+	info.Summary = "Detects suspicious sort.Slice calls"
+	info.Before = `sort.Slice(xs, func(i, j) bool { return keys[i] < keys[j] })`
+	info.After = `sort.Slice(kv, func(i, j) bool { return kv[i].key < kv[j].key })`
+
+	collection.AddChecker(&info, func(ctx *lintpack.CheckerContext) lintpack.FileWalker {
+		return astwalk.WalkerForExpr(&sortSliceChecker{ctx: ctx})
+	})
+}
+
+type sortSliceChecker struct {
+	astwalk.WalkHandler
+	ctx *lintpack.CheckerContext
+}
+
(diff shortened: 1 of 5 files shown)


checkers/sortSlice_checker.go
```diff
@@ -0,0 +1,135 @@
...
+		// or misprint in sorting condition.
+		// We give a warn if X contains indexing with `i` index and Y
+		// contains indexing with `j`.
+		if c.containsIndex(cmp.X, jvar) && c.containsIndex(cmp.Y, ivar) {
+			c.warnIndex(cmp, ivar, jvar)
+		}
+	}
+}
+
+func (c *sortSliceChecker) paramIdents(e *ast.FuncType) (*ast.Ident, *ast.Ident) {
```
cristaloleg, line 90: ?
```
fields := e.Params.List
if len(fields) == 2 {
    return fields.Names[0], fields.Names[1]
}
return nil, nil
```

checkers/testdata/sortSlice/negative_tests.go
```diff
@@ -0,0 +1,106 @@
+package checker_test
+
+import (
+	"sort"
+)
+
+func goodSorting() {
+	{
```
cristaloleg, line 8: I think we should not support `f(x)`, isn't it?
```
var xs []int
sort.Slice(ys, func(i, j int) bool {
    return foo(ys[i]) >= foo(ys[j])
})
```

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/go-toolsmith/astfmt v1.0.0
 	github.com/go-toolsmith/astp v1.0.0
 	github.com/go-toolsmith/strparse v1.0.0
-	github.com/go-toolsmith/typep v1.0.0
+	github.com/go-toolsmith/typep v1.0.2
```
cristaloleg, line 13: <3

checkers/sortSlice_checker.go
quasilyte: This won't handle the case where both idents are a single field.
```
i, j int - 1 field with 2 names
i int, j int - 2 fields, each with 1 name
```

cristaloleg: oh! got it, nice.

checkers/testdata/sortSlice/negative_tests.go
quasilyte: This case should not give any warnings in the current implementation due to several checks:
1. return expression contains side effects (function calls).
2. there are `ys` inside both side of the `>=` operator.

We can't give a warning for that case since there are valid cases that involve key generation based on the element.

Imagine that we want to sort ints using string-like order:
```
var xs []int
sort.Slice(ys, func(i, j int) bool {
    return int2string(ys[i]) >= int2string(ys[j])
})
```
The only case we're trying to catch is using conditions that do not depend on the slice being sorted.

cristaloleg: My idea was to add such test (skipping func calls), unless I've missed it.

quasilyte: There is this:
image

But I can add another one with slices in function calls.

quasilyte: Done.

quasilyte:
PTAL.

PR #930, opened 2020-05-26
checkers: add filepath.Join checker

quasilyte:
```
Reports filepath.Join() string literal arguments that contain
'/' or '\' as it kinda defeats the purpose of using that function.

```

### DIFF
--- a/checkers/testdata/filepathJoin/negative_tests.go
+++ b/checkers/testdata/filepathJoin/negative_tests.go
@@ -1,0 +1,16 @@
+package checker_test
+
+import (
+	"path/filepath"
+)
+
+func badArgs() {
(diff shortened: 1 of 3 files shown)


checkers/testdata/filepathJoin/negative_tests.go
```diff
@@ -0,0 +1,16 @@
+package checker_test
+
+import (
+	"path/filepath"
+)
+
+func badArgs() {
```
cristaloleg, line 7: are they really bad?

quasilyte: They could be better. :sweat_smile:

PR #932, opened 2020-05-26
checkers: add sqlQuery checker

quasilyte:
Reports `_, err := db.Query()` pattern.
Suggests to use Exec() instead of Query in this case.

### DIFF
--- a/checkers/sqlQuery_checker.go
+++ b/checkers/sqlQuery_checker.go
@@ -1,0 +1,164 @@
+package checkers
+
+import (
+	"go/ast"
+	"go/types"
+
+	"github.com/go-lintpack/lintpack"
+	"github.com/go-lintpack/lintpack/astwalk"
+	"github.com/go-toolsmith/astcast"
+)
+
+func init() {
+	var info lintpack.CheckerInfo
+	info.Name = "sqlQuery"
+	info.Tags = []string{"diagnostic", "experimental"}
+	info.Summary = "Detects issue in Query() and Exec() calls"
+	info.Before = `_, err := db.Query("UPDATE ...")`
+	info.After = `_, err := db.Exec("UPDATE ...")`
+
+	collection.AddChecker(&info, func(ctx *lintpack.CheckerContext) lintpack.FileWalker {
+		return astwalk.WalkerForStmt(&sqlQueryChecker{ctx: ctx})
+	})
+}
+
+type sqlQueryChecker struct {
+	astwalk.WalkHandler
+	ctx *lintpack.CheckerContext
+}
+
+func (c *sqlQueryChecker) VisitStmt(stmt ast.Stmt) {
+	assign := astcast.ToAssignStmt(stmt)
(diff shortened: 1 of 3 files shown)


checkers/sqlQuery_checker.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+	}
+	return false
+}
+
+func (c *sqlQueryChecker) funcIsQuery(funcExpr *ast.SelectorExpr) bool {
+	switch funcExpr.Sel.Name {
+	case "Query":
+		// Stdlib and friends.
+	case "Queryx", "QueryRowx":
```
un000, line 76: QueryRowx is redundant, because a returned value with a type Row hasn't Close() method

Also I suggest to add this methods:
```
QueryContext
QueryxContext
```

quasilyte: Removed QueryRowx.
Added QueryContext and QueryxContext.

checkers/sqlQuery_checker.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	if c.typeHasExecMethod(c.ctx.TypesInfo.TypeOf(funcExpr.X)) {
+		c.warnAndSuggestExec(funcExpr)
+	} else {
+		c.warnRowsIgnored(funcExpr)
+	}
+}
+
+func (c *sqlQueryChecker) typeIsRowsLike(typ types.Type) bool {
```
cristaloleg, line 59: move down? after `funcIsQuery`

checkers/sqlQuery_checker.go
```diff
@@ -0,0 +1,151 @@
...
+	switch typ := typ.(type) {
+	case *types.Pointer:
+		return c.typeIsRowsLike(typ.Elem())
+	case *types.Named:
+		if typ.Obj().Name() == "Rows" {
+			return true
+		}
+	}
+	return false
```
cristaloleg, line 68: wdyt?
```suggestion
		return if typ.Obj().Name() == "Rows"
	default:
		return false
	}
	
```

checkers/sqlQuery_checker.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func (c *sqlQueryChecker) funcIsQuery(funcExpr *ast.SelectorExpr) bool {
+	switch funcExpr.Sel.Name {
+	case "Query", "QueryContext":
+		// Stdlib and friends.
+	case "Queryx", "QueryxContext":
+		// sqlx.
```
cristaloleg, line 76: <3

checkers/sqlQuery_checker.go
quasilyte: Fixed.

checkers/sqlQuery_checker.go
quasilyte: Done.

quasilyte:
Added aliases test. PR needs approval again.

Issue #854, opened 2019-05-09
Consider transition to tools/go/analysis


quasilyte:
See https://godoc.org/golang.org/x/tools/go/analysis.

quasilyte:
I experimented with `analysis` and believe that we can do that transition. 
Especially since we still do only file-based linting.

cristaloleg:
Looks like 1, 4 is left (where one is a minor thing).

quasilyte:
I did some more digging. We can do this transition relatively easy if we get rid of the shared `Context` thing.
`CheckerContext` could include everything we need.

The only problem is our `require` system, but it can be, in theory, replaced with go/analysis based pass-dependency system.

cristaloleg:
Sounds good

PR #937, opened 2020-06-15
all: integrate lintpack as internal framework

quasilyte:
Instead of using a lintpack as a dependency, clone
most of its code that we need.

This is an intermediate step to make transition to
go/analysis easier. Now we don't need to keep
lintpack dependency in sync and can make
framework/linter compatible with go/analysis later.

Doing a v0.5.0 release is suggested.

Updates #854


Makefile
```diff
@@ -40,7 +40,7 @@ cover:
 	goveralls -package github.com/go-critic/go-critic/checkers -coverprofile=coverage.out -service travis-ci -repotoken ${COVERALLS_TOKEN}
 
 gocritic:
-	lintpack build -o gocritic -linter.version='v0.4.3' -linter.name='gocritic' github.com/go-critic/go-critic/checkers
+	go build -o gocritic ./cmd/gocritic
```
cristaloleg, line 43: <3

PR #947, opened 2020-08-06
Allow []string as config value type

Eun:
I propose the feature of using string slices as config value types.

My use case: I am using `ruleguard` and want to provide a set of multiple rule files.

This PR implements the `StringSlice()` function on `CheckerParams`.

In additon it adds some `IsType()` functions, to determinate what type the config value is, e.g. (`IsInt()`, `IsBool()`,...)

It also adds the possibility to use multiple rule files for ruleguard. (my usecase).

However I am unsure howto add a correct integration test. And I am unsure if slices can be passed as a parameter.

So definitely need help for this.


quasilyte:
Hi! Thank you for the PR.
I think having an ability to specify multiple rules files is useful.
The ruleguard standalone binary actually supports it, so it should also be supported by the go-critic.

I'll take a look into the code today or tomorrow. Overall, it looks good, but I want to be sure we don't miss some details.

quasilyte:
I also want to hear @cristaloleg opinion on this.

checkers/ruleguard_checker.go
```diff
@@ -34,62 +35,89 @@ func init() {
...
 		return c
 	}
 
+	for _, rulesFileName := range rules {
+		if rulesFileName == "" {
+			continue
+		}
+		rset, err := parseRuleGuardFile(rulesFileName)
+		if err != nil {
+			log.Printf("unable to parse file %s: %+v", rulesFileName, err)
```
cristaloleg, line 54: Probably we should inform that other files will be skipped.

checkers/ruleguard_checker.go
```diff
@@ -34,62 +35,89 @@ func init() {
...
 	// a transition to the go/analysis framework.
 	//
 	// For now, we log error messages and return a ruleguard checker
 	// with an empty rules set.
-
 	data, err := ioutil.ReadFile(rulesFilename)
 	if err != nil {
-		log.Printf("ruleguard init error: %+v", err)
-		return c
+		return nil, fmt.Errorf("ruleguard init error: %+v", err)
```
cristaloleg, line 73: looks like a good place for %w :)

checkers/ruleguard_checker.go
```diff
@@ -34,62 +35,89 @@ func init() {
...
 type ruleguardChecker struct {
 	ctx *linter.CheckerContext
-
-	rset *ruleguard.GoRuleSet
+	rset []goRuleSetWithFileName
 }
 
 func (c *ruleguardChecker) WalkFile(f *ast.File) {
-	if c.rset == nil {
+	if len(c.rset) == 0 {
```
cristaloleg, line 95: Nice!

checkers/ruleguard_checker.go
Eun: You are right, I wasn't sure if go-critic uses go 1.13

quasilyte:
What about this: keep `rules` param as string and do a split on comma.
So we can treat as both list and a single filename.

quasilyte:
Instead of running all rule sets one by one, we can use a new MergeRuleSets function.
I also preserved the rule defining filename as it looks like you're interested in it.
See https://github.com/quasilyte/go-ruleguard/pull/64

Eun:
Nice! Thanks @quasilyte

quasilyte:
We updated ruleguard and now it's possible to use  the ruleset merging function. :)
https://github.com/go-critic/go-critic/pull/952

quasilyte:
I think we should come back to this and enable multiple rule files for the ruleguard. It should be easier now that we have a merging function.

PR #949, opened 2020-08-09
Enable govet in golangci config (#948)

### LINKED ISSUE #948: Enable govet in golangci config
We disabled go vet at some point, but we probably want to enable it back.

LevOspennikov:
resolve #948 
Hi guys!
Can't find reason while it was disabled, so just recover this string from old make config


quasilyte:
I think it's easier to use go vet from the golangci-lint?
We need to remove govet from the disabled list (see `.golangci.yml`).

PR #958, opened 2020-08-16
checkers: use nil-safe TypeOf

quasilyte:
Added ruleguard rule to find all current cases and
to prevent them from showing up again in future.


quasilyte:
The first build is expected to fail as I haven't fixed anything yet.
I want to check that ruleguard is executed.

Update: whoops. For some reason it doesn't...
Maybe we need to update it?

Issue #488, opened 2018-07-26
lint: file-scoped check that requires type definition before its method definitions


quasilyte:
Opinionated check. :nerd_face:
File-scoped.

Detects method definitions that go before receiver type definition.

**Before:**
```go
func (r recv) Method() {}
type recv struct{}
```

**After:**
```go
type recv struct{}
func (r recv) Method() {}
```

Quasilyte:
We're lacking FileVisitor and/or DeclVisitor that would be a good fit for this.
I'll fix that soon.

Quasilyte:
The implementation can look like this:
Walk file declarations, looking and every `*ast.FuncDecl` and `*ast.GenDecl` with `token.TYPE` kind.
If we have a function with receiver of type `T` that was not previously been defined with `*ast.GenDecl`, issue a warning and mark `T` as one that already caused warning (to avoid duplicated warnings). One may use type name as a map key or use `[]types.Type` list and use `types.Identical` function to check for equality.

PR #960, opened 2020-08-16
#488 type def first checker

LevOspennikov:
Add checker, that requires to declare the definition of type before its methods (#488)


checkers/typeDefFirst_checker.go
```diff
@@ -0,0 +1,75 @@
...
+		switch decl := declaration.(type) {
+		case *ast.FuncDecl:
+			if decl.Recv != nil {
+				receiver := decl.Recv.List[0]
+				typeName := trimAsterisk(NodeToString(t.ctx.FileSet, receiver.Type))
+				typeUsageMap[typeName] = true
+			}
+
+		case *ast.GenDecl:
+			if decl.Tok == token.TYPE {
```
LevOspennikov, line 54: can reduce nesting, but make harder control flow, don't know, which way is better

checkers/typeDefFirst_checker.go
```diff
@@ -0,0 +1,75 @@
...
+
+	if f.Decls == nil {
+		return
+	}
+
+	for _, declaration := range f.Decls {
+		switch decl := declaration.(type) {
+		case *ast.FuncDecl:
+			if decl.Recv != nil {
+				receiver := decl.Recv.List[0]
```
LevOspennikov, line 48: should I check the array for not nilness? as far as I understood, in this case (where its receiver) there are always should be some fields, but the safer the better?

checkers/utils.go
```diff
@@ -307,3 +309,7 @@ func identOf(x ast.Node) *ast.Ident {
 		return nil
 	}
 }
+
+func NodeToString(set *token.FileSet, node ast.Node) string {
```
LevOspennikov, line 313: if the project already has similar function plz, point my nose

checkers/typeDefFirst_checker.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/go-critic/go-critic/checkers/internal/astwalk"
+	"github.com/go-critic/go-critic/framework/linter"
+)
+
+func init() {
+	var info linter.CheckerInfo
+	info.Name = "typeDefFirst"
+	info.Tags = []string{"style", "experimental"}
+	info.Summary = "File-scoped checker, that requires type definition before its method definitions"
```
cristaloleg, line 16: `File-scoped checker` I'm curious - should be use this phrase in other checkers? maybe another tag or subtype? @quasilyte :)

checkers/typeDefFirst_checker.go
```diff
@@ -0,0 +1,75 @@
...
+	})
+}
+
+type typeDefFirstChecker struct {
+	astwalk.WalkHandler
+	ctx *linter.CheckerContext
+}
+
+func (t *typeDefFirstChecker) WalkFile(f *ast.File) {
+	typeUsageMap := make(map[string]bool)
```
cristaloleg, line 38: can be moved lower

checkers/typeDefFirst_checker.go
cristaloleg: you can add a simple `func` in golden files if it fails - better to add :)

checkers/typeDefFirst_checker.go
cristaloleg: it's ok to leave it as it is.

checkers/typeDefFirst_checker.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+}
+
+func (t *typeDefFirstChecker) warn(cause ast.Node, typeName string) {
+	t.ctx.Warn(cause, "definition of type '%s' should appear before its methods", typeName)
+}
+
+// removing '*' from type
+func trimAsterisk(typeName string) string {
+	return strings.TrimLeft(typeName, "* ")
```
cristaloleg, line 74: huh, space?

checkers/utils.go
```diff
@@ -1,7 +1,9 @@
 package checkers
 
 import (
+	"github.com/go-toolsmith/astfmt"
```
cristaloleg, line 4: move it after `"github.com/go-critic/go-critic/framework/linter"` please

checkers/utils.go
cristaloleg: I'm not 100% sure that we need this func yet 👀

quasilyte:
I can give a review when Oleg comments are addressed. :)

checkers/typeDefFirst_checker.go
LevOspennikov: It does not fail, because it goes after `decl.Recv != nil` where Recv is recivier

checkers/utils.go
LevOspennikov: so, what should we do?

checkers/typeDefFirst_checker.go
LevOspennikov: yeah, fair enough

checkers/typeDefFirst_checker.go
LevOspennikov: It is in case of (typeName * type), but it works without space too, unexpectedly. Removed

checkers/utils.go
quasilyte: AFAIK, the receiver name node is either `*ast.Ident` or `*ast.StarExpr`. So instead of doing an AST->string->trimAsterisk you can do:
```go
func (c *typeDefFirstChecker) receiverType(e ast.Expr) string {
  switch e := e.(type) {
  case *ast.Pointer:
    return c.receiverType(e.X)
  case *ast.Ident:
    return e.Value
  default:
    panic("unreachable")
  }
}
```
The `trimAsterisk` function, if we even wanted it, should be a method of the checker type. All helpers that are not generally useful should be encapsulated as a checker detail. As a general rule: no top-level functions should be defined inside the checker file.

checkers/typeDefFirst_checker.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+	})
+}
+
+type typeDefFirstChecker struct {
+	astwalk.WalkHandler
+	ctx *linter.CheckerContext
+}
+
+func (t *typeDefFirstChecker) WalkFile(f *ast.File) {
```
quasilyte, line 37: We have a convention to have a `c` for every checker receiver. This makes checker renaming easier (and we do rename experimental checkers from time to time before they get stable). It also makes the code more consistent.

checkers/typeDefFirst_checker.go
quasilyte: I would actually prefer having something like this:
```go
if decl.Tok != toke.Type { 
  continue
}
// Happy path without indentation
```
I'm not sure it makes the control flow harder.

checkers/typeDefFirst_checker.go
```diff
@@ -0,0 +1,75 @@
...
+				receiver := decl.Recv.List[0]
+				typeName := trimAsterisk(NodeToString(t.ctx.FileSet, receiver.Type))
+				typeUsageMap[typeName] = true
+			}
+
+		case *ast.GenDecl:
+			if decl.Tok == token.TYPE {
+				for _, spec := range decl.Specs {
+					if spec, ok := spec.(*ast.TypeSpec); ok {
+						typeName := trimAsterisk(spec.Name.Name)
```
quasilyte, line 57: If I remember correctly, `*ast.Ident` (`spec.Name.Name`) can never have `*` in it's value.
So using `spec.Name.Name` should be enough.

checkers/typeDefFirst_checker.go
```diff
@@ -0,0 +1,75 @@
...
+			if decl.Recv != nil {
+				receiver := decl.Recv.List[0]
+				typeName := trimAsterisk(NodeToString(t.ctx.FileSet, receiver.Type))
+				typeUsageMap[typeName] = true
+			}
+
+		case *ast.GenDecl:
+			if decl.Tok == token.TYPE {
+				for _, spec := range decl.Specs {
+					if spec, ok := spec.(*ast.TypeSpec); ok {
```
quasilyte, line 56: You can decrease the nesting even further:
```go
for _, spec := range decl.Specs {
	spec, ok := spec.(*ast.TypeSpec)
	if !ok {
		continue
	}
```
This idiom is quite common.

quasilyte:
The algorithm looks good.

NodeToString function can be either replaced with the proposed type switch, or (perhaps this is preferred), there is a `astfmt.Sprint(node)` which already looks convenient enough.

checkers/typeDefFirst_checker.go
LevOspennikov: I know, but `continue` in nested for it is a little harder to understand, that is why I asked

checkers/utils.go
LevOspennikov: Wow, I am not familiar with all type relations, maybe some map or guide exists? That makes `trimAsterisk` 
unnecessary

LevOspennikov:
Thanks for the detailed review, updated request, using type switch now.

checkers/utils.go
quasilyte: I think the closest thing is Go spec.
Although it's not always the most precise source as it defines valid receiver in its grammar as a normal param, but you can't have a slice type for a receiver, for example.

checkers/typeDefFirst_checker.go
quasilyte: You know the solution to that: don't nest a lot of fors. :)
It's always possible to define another method and call that, so you don't have too much control flow on the same level.
But I'm OK with the current level of nesting. AST requires a lot of things like this, unfortunately.

quasilyte:
I'm almost positive that we'll need to figure out a better name for this checker. But it's OK for now as it's marked as experimental.

PR #961, opened 2020-08-16
all: enable linter on CI + corresponding fixes

cristaloleg:


rules.go
```diff
@@ -0,0 +1,11 @@
+package gorules
+
+import (
+	"github.com/quasilyte/go-ruleguard/dsl/fluent"
+)
+
+//nolint:deadcode,unused
```
quasilyte, line 7: I propose adding `//nolint` to the line above `package gorules`. It will disable linting for this file completely.

PR #962, opened 2020-08-16
checkers: minor refactoring of typeDefFirst checker

quasilyte:
Follow-up for the #960

### DIFF
--- a/checkers/typeDefFirst_checker.go
+++ b/checkers/typeDefFirst_checker.go
@@ -12,7 +12,7 @@ func init() {
 	var info linter.CheckerInfo
 	info.Name = "typeDefFirst"
 	info.Tags = []string{"style", "experimental"}
-	info.Summary = "File-scoped checker, that requires type definition before its method definitions"
+	info.Summary = "Detects method declarations preceding the type definition itself"
 	info.Before = `
 func (r rec) Method() {}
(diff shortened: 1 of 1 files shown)


quasilyte:
CC @LevOspennikov

checkers/typeDefFirst_checker.go
```diff
@@ -30,44 +30,48 @@ func (r rec) Method() {}
 
 type typeDefFirstChecker struct {
 	astwalk.WalkHandler
-	ctx *linter.CheckerContext
+	ctx          *linter.CheckerContext
+	trackedTypes map[string]bool
```
cristaloleg, line 34: Weeeell, in the end it can be map-string-struct :)

LevOspennikov, line 34: nit: why do we need this parameter? Feels like it can be the second argument of the function `walkDecl()`, it makes function purer, idk.

quasilyte: But it's not a function, it's a method.
And trackedTypes is not a parameter, it's a struct field.
Checkers are generally can't be used without a state, they usually have at least some kind of state.

cristaloleg: I'm ok with a field

PR #970, opened 2020-09-11
checkers: add builtinShadowDecl

### LINKED ISSUE #868: gocritic print nothing in case of builtinShadow for int type
```
cat main.go
package main

import "fmt"

type int struct{}

func main() {
  var i int
  _ = i
  fmt.Println(i)
}
```

`$ ./gocritic check -enable builtinShadow main.go`
`./gocritic version v0.3.4`

quasilyte:
Finds top-level symbols that have a name that will shadow a
builtin identifier for the entire package.

Unlike builtinShadow, it's not a style but rather a diganostic check.

Fixes #868

### DIFF
--- a/checkers/testdata/builtinShadowDecl/positive_tests.go
+++ b/checkers/testdata/builtinShadowDecl/positive_tests.go
@@ -1,0 +1,28 @@
+package checker_test
+
+/*! shadowing of predeclared identifier: int */
+type int struct{}
+
+type (
+	/*! shadowing of predeclared identifier: int8 */
+	int8 = int
+	/*! shadowing of predeclared identifier: int16 */
+	int16 = int
+)
+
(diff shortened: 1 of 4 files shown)


checkers/testdata/builtinShadowDecl/positive_tests.go
```diff
@@ -0,0 +1,21 @@
+package checker_test
```
cristaloleg, line 1: Can you add `type int = struct{}` too?

cristaloleg (dismissed):

checkers/testdata/builtinShadowDecl/positive_tests.go
quasilyte: Done.

PR #984, opened 2020-09-20
framework/linttest: fail on pkg errors by default

### LINKED ISSUE #977: dev: add makefile step to compile test files
Sometimes we can fall into the situation when positive/negative files cannot be compiled due to mistakes. Sometimes it's hard to catch. Automation can make it simpler.

quasilyte:
If a package loaded from the testdata contains errors,
fail the test unless there is an entry in IgnoreErrors slice
that asks the linttest to skip error checking for that checker.

Fixes #977

### DIFF
--- a/checkers/checkers_test.go
+++ b/checkers/checkers_test.go
@@ -22,7 +22,13 @@ func TestCheckers(t *testing.T) {
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


cristaloleg (approved):
WoAh

PR #993, opened 2020-10-04
dev: added makefile step to generate new checker

### LINKED ISSUE #976: dev: makefile step to init new checker
Makefile step should create simple checker with a given name and directory with the empty test files.

neverix:
Fixes #976.

Implemented using plain Makefile. It might be a better idea to write a small shell script or go program to avoid using GNU Make-specific hacks.


cristaloleg (approved):
<3
I'm not 100% sure that we need to configure checker's type via Makefile and defaults should be fine, but will leave this question to Iskander.

cristaloleg:
I think it's time to merge it. Sorry, no idea why it was so long under our radar.

neverix:
I totally forgot about this

PR #1005, opened 2020-11-20
checkers: extend flagName diagnostics

quasilyte:
From the https://github.com/golang/go/issues/41792

    And these are invalid flag names:

        * Empty String:

          * If the flag name is an empty string and is used in the `-flag x` form, the flag starting with hyphen('-''') can be the non-flag argument and the flag starting with double hyphen('--''') can be the terminator.
          * > Flag parsing stops just before the first non-flag argument("-" is a non-flag argument) or after the terminator "--".

        * Starting with hyphen:

          * According to the document the flag should start with '-' or '--'. But, if the flag name starts with hyphen, this package can’t distinguish between '-''-flag' and '--''flag'.

        * Containing equal sign:

          * This package uses equal sign to split the flag into a name-value pair. So, if the flag name contains equal sign, this package can't split it properly.
          * e.g. If the flag name is 'foo=bar' and the flag is '-foo=bar=value', this package can't find out the actual value.

### DIFF
--- a/checkers/flagName_checker.go
+++ b/checkers/flagName_checker.go
@@ -15,9 +15,10 @@ func init() {
 	var info linter.CheckerInfo
 	info.Name = "flagName"
 	info.Tags = []string{"diagnostic"}
-	info.Summary = "Detects flag names with whitespace"
+	info.Summary = "Detects suspicious flag names"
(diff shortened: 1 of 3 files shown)


cristaloleg (approved):
🚀

PR #1020, opened 2021-01-26
Checker constructor may return error

sebastien-rosset:
The purpose of this PR is to provide a cleaner way to handle checker initialization issues. For example, when `ruleguard` parses input files, it may encounter syntax errors, in which case the checker constructor should return an error, which can be propagated to the callers, including **golangci-lint**.

This is done by changing the signature of the following functions:
```diff
- func (coll *CheckerCollection) AddChecker(info *CheckerInfo, constructor func(*CheckerContext) FileWalker)
+ func (coll *CheckerCollection) AddChecker(info *CheckerInfo, constructor func(*CheckerContext) (FileWalker, error)) 

- func NewChecker(ctx *Context, info *CheckerInfo) *Checker {
+ func NewChecker(ctx *Context, info *CheckerInfo) (*Checker, error) 
```

I'm not sure if it's ok to change the signature of these exported functions. It seems somewhat unlikely they are invoked outside this module, but I don't have enough visibility into the usage.

Related PR in golangci-lint: https://github.com/golangci/golangci-lint/pull/1666


sebastien-rosset:
@quasilyte , @cristaloleg

cristaloleg (approved):
🎉

sebastien-rosset:
If this PR is merged, maybe we don't need the `failOnError` setting anymore. It would make sense to always exit with error if the ruleguard checker encounters an error during initialization.

quasilyte (approved):

PR #1024, opened 2021-02-23
granular error handling

sebastien-rosset:


quasilyte:
Looks like a good enough solution for now.
Let's update the ruleguard dependency version to `v0.3.2` and merge it.

@cristaloleg, what's your opinion on this?

sebastien-rosset:
WIP

quasilyte:
OK. Let me know when it's ready for review. :+1:

sebastien-rosset:
> OK. Let me know when it's ready for review. 👍

Sure, I am adding more unit tests.

sebastien-rosset:
@quasilyte I need some help for linttest, which is used in the integration tests.
Here: https://github.com/go-critic/go-critic/blob/master/framework/linttest/linttest.go#L25
and here: https://github.com/go-critic/go-critic/blob/a5c208f62c4bbc29272fcf4140e76e663e7e39d8/checkers/checkers_test.go#L36

Shouldn't this be a path relative to the working directory?
If you fork go-critic to submit a bugfix/enhancement, doesn't that mean during the tests you will still run the upstream code?

quasilyte:
So far I've found that we don't actually return an import error from the ruleguard.
We always return `parseError`.
Here is a fix: https://github.com/quasilyte/go-ruleguard/pull/220

quasilyte:
With `v0.3.3`, import errors should be distinguishable.

sebastien-rosset:
@quasilyte could you review? Initially I tried to add integration tests, but the integration tests only seem to work with the original repo, not a fork.

quasilyte:
It took some time, but we're here at last.

PR #1031, opened 2021-03-15
checkers: commentFormatting: allow `-` in pragma tag names

### LINKED ISSUE #1027: commentFormatting complains about some directives
```go
//go-sumtype:decl Data
//go:noinline
```

The first one is used for https://github.com/BurntSushi/go-sumtype.

See also https://github.com/mvdan/gofumpt/issues/109 and https://github.com/mvdan/gofumpt/blob/d7cd0a5d35ce5779a01c803173a92769e330c2d1/format/format.go#L213-L225

quasilyte:
Fixes #1027

### DIFF
--- a/checkers/commentFormatting_checker.go
+++ b/checkers/commentFormatting_checker.go
@@ -22,7 +22,7 @@ func init() {
 	collection.AddChecker(&info, func(ctx *linter.CheckerContext) (linter.FileWalker, error) {
 		parts := []string{
 			`^//go:generate .*$`, // e.g.: go:generate value
-			`^//\w+:.*$`,         // e.g.: key: value
+			`^//[\w-]+:.*$`,      // e.g.: key: value
 			`^//nolint\b`,        // e.g.: nolint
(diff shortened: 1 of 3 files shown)


checkers/commentFormatting_checker.go
```diff
@@ -74,6 +74,6 @@ func (c *commentFormattingChecker) specialChar(r rune) bool {
 	}
 }
 
-func (c *commentFormattingChecker) warn(cg *ast.CommentGroup) {
-	c.ctx.Warn(cg, "put a space between `//` and comment text")
+func (c *commentFormattingChecker) warn(comment *ast.Comment) {
```
cristaloleg, line 77: why so? :)

cristaloleg: Looks like a bug made before, but not sure.

cristaloleg (approved):

checkers/commentFormatting_checker.go
quasilyte: Previously, it would complain on the beginning of the comment block.
Now it should report the exact comment line instead.

PR #1034, opened 2021-03-26
Fix appendAssign checker to also warn on short decl

adlerjohn:
Makes the `appendAssign` checker also warn on a short declaration, not just on assignments, e.g.

```go
x := []int{}
y := append(x, 1) // <- bad!
```

The motivation for this restriction is that allowing short declarations is error prone: https://blog.allegro.tech/2017/07/golang-slices-gotcha.html

Additional context: https://stackoverflow.com/a/35276346


quasilyte (approved):

quasilyte:
Thank you for the contribution!

PR #1043, opened 2021-04-28
Add `preferDecodeRune`

tdakkota:
This PR adds a simple checker to find `[]rune("abc")[0]` and suggest `utf8.DecodeRuneInString("abc")`. 
Expressions like 
```go
[]rune("abc")[0]
```
causes unnecessary rune slice allocation.

See https://github.com/golang/go/issues/45260.

### DIFF
--- a/checkers/preferDecodeRune_checker.go
+++ b/checkers/preferDecodeRune_checker.go
@@ -14,7 +14,7 @@ import (
 func init() {
 	var info linter.CheckerInfo
 	info.Name = "preferDecodeRune"
-	info.Tags = []string{"performance"}
+	info.Tags = []string{"performance", "experimental"}
 	info.Summary = "Detects expressions like []rune(s)[0] that may cause unwanted rune slice allocation"
 	info.Before = `r := []rune(s)[0]`
 	info.After = `r, _ := utf8.DecodeRuneInString(s)`
--- a/checkers/preferDecodeRune_checker.go
+++ b/checkers/preferDecodeRune_checker.go
@@ -1,0 +1,56 @@
+package checkers
+
+import (
+	"go/ast"
+	"go/types"
+
+	"github.com/go-toolsmith/astcast"
+	"github.com/go-toolsmith/typep"
+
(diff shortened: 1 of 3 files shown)


checkers/preferDecodeRune_checker.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/go-toolsmith/typep"
+
+	"github.com/go-critic/go-critic/checkers/internal/astwalk"
+	"github.com/go-critic/go-critic/framework/linter"
+)
+
+func init() {
+	var info linter.CheckerInfo
+	info.Name = "preferDecodeRune"
+	info.Tags = []string{"performance"}
```
quasilyte, line 17: Please add `"experimental"` tag as all new checkers should stay experimental for at least 1 release; so we can test them while affecting less users with potential bugs.

quasilyte:
Otherwise, the code looks very nice!

checkers/preferDecodeRune_checker.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+type preferDecodeRuneChecker struct {
+	astwalk.WalkHandler
+	ctx *linter.CheckerContext
+}
+
+func (c *preferDecodeRuneChecker) VisitExpr(e ast.Expr) {
+	indexExpr := astcast.ToIndexExpr(e)
+	indexed := indexExpr.X
```
cristaloleg, line 35: This can be `nil` afair. Will next line fail? 👀

cristaloleg (approved):
LGTM, thank you <3

checkers/preferDecodeRune_checker.go
quasilyte: AFAIK, `TypeOf(nil)` is safe.
And astcast returns a sentinel object in case of a failed case, so it's safe to access its fields.

cristaloleg (approved):

quasilyte (approved):

quasilyte:
When we'll merge #1044, it should be possible to rewrite this checker as a ruleguard rule.

PR #1045, opened 2021-04-30
checkers: rewrite preferDecodeRune as ruleguard rule

quasilyte:


quasilyte:
@tdakkota what do you think?

PR #1048, opened 2021-05-06
Fixed deferUnlambda `After` docs `f()`->`defer f()`

pregnor:
Fixed the `deferUnlambda` checker's `After` documentation string `f()`->`defer f()`, to indicate the checker doesn't expect the removal of the defer statement, only expects to transform the lambda wrapped function call into a simple function call.

This is in line with the current functionality which [expects](https://github.com/go-critic/go-critic/blob/6090d88757f445be96d2e507d7b17e2abcb40c4b/checkers/deferUnlambda_checker.go#L93) the corrected code to look like `defer f()`.


cristaloleg (approved):
Nice catch! Thank you!

quasilyte (approved):
:tada:

PR #1067, opened 2021-08-02
checkers: add preferWriteByte

tdakkota:


cristaloleg (approved):
LGTM, thx

quasilyte:
Negative tests are tests without warnings; negative - no triggers.
Positive tests are tests with warnings; positive - have triggers.
The naming choice could be confusing, but this is why we can give a piece of advice here, on a review. :)

Also, I would like to see a negative test where WriteRune is called on a type that doesn't implement byte writer.

cristaloleg (approved):

quasilyte (approved):

PR #1080, opened 2021-10-09
dupArg: rewrite with ruleguard

cristaloleg:


quasilyte:
The test failure is expected and intended to show the problem.
It should go away when we update ruleguard.
https://github.com/quasilyte/go-ruleguard/pull/260

PR #1108, opened 2021-10-13
checkers: add unassignedError

cristaloleg:


un000 (dismissed):

quasilyte:
This needs to be rebased and re-generated.
After that, we can merge it.

quasilyte:
Now we need to check the warnings for the external testdata.
If they're expected, they need to be added to the golden files (PR to the external testdata repo).

PR #1111, opened 2021-10-14
checkers: add exposedSyncMutex

cristaloleg:


AlekSi:
I wonder how that would work with types implementing
```go
interface {
  Lock()
  Unlock()
}
```

and `copylock` vet check. It might be worth adding to tests.

PR #1127, opened 2021-10-18
ci: try other OS

cristaloleg:


quasilyte:
By the way.
In order to reproduce the issue, our CI needs to download the release binary.
Compiling gocritic from the source won't reveal any failure.

quasilyte:
Is this still relevant?

cristaloleg:
Not so, probably we will solve this with goreleaser

Issue #1135, opened 2021-10-19
go-toolsmith/pkgload@v1.0.0 panics in some modules


MichaelUrman:
Testing `@v0.6.0` and `@latest` I encountered a panic in pkgload.

I found go-toolsmith/pkgload#9 addresses this. After cloning master, running `go get github.com/go-toolsmith/pkgload@v1.0.1`, and installing ./cmd/gocritic, I no longer encounter the panic when checking the same module. It appears to be specific to something in this module, as checking other modules did not panic.

```
% go install go install github.com/go-critic/go-critic/cmd/gocritic@v0.6.0
% gocritic check ./...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x8912a0]

goroutine 1 [running]:
github.com/go-toolsmith/pkgload.VisitUnits.func4(0xc000db98a0, 0xc007236b00)
	/home/murman/go/pkg/mod/github.com/go-toolsmith/pkgload@v1.0.0/pkgload.go:125 +0x40
sort.doPivot_func({0xc000db98a0, 0xc007236b00}, 0x0, 0x4f)
	/usr/local/go/src/sort/zfuncversion.go:87 +0x1d9
sort.quickSort_func({0xc000db98a0, 0xc007236b00}, 0xc0009e7fe0, 0x0, 0xc000db98e0)
	/usr/local/go/src/sort/zfuncversion.go:143 +0x85
sort.Slice({0x8b9d20, 0xc0009e7fe0}, 0x4f)
	/usr/local/go/src/sort/slice.go:20 +0x9f
github.com/go-toolsmith/pkgload.VisitUnits({0xc008af9400, 0xc0000200a0, 0x1}, 0xc000db9a70)
	/home/murman/go/pkg/mod/github.com/go-toolsmith/pkgload@v1.0.0/pkgload.go:124 +0x6e9
github.com/go-critic/go-critic/framework/lintmain/internal/check.loadPackages(0x8d6260, {0xc0000200a0, 0x930108, 0x5})
	/home/murman/go/pkg/mod/github.com/go-critic/go-critic@v0.6.0/framework/lintmain/internal/check/check.go:471 +0x74
github.com/go-critic/go-critic/framework/lintmain/internal/check.(*program).loadProgram(0xc0005c4000)
[...]

PR #1136, opened 2021-10-20
update go-toolsmith/pkgload

MichaelUrman:
This avoids a nil deref in some modules.


MichaelUrman:
Fixes #1135 

As alluded to there, I have not isolated the underlying problem so have not added a test case here. I have also not evaluated other dependencies for upgrades. Several are available, including google/go-cmp, logrusorgru/aurora, mattn/goveralls, phorman/uuid, quasilyte/go-consistent, quasilyte/go-ruleguard, quasilyte/regex, and x/tools.

PR #1156, opened 2021-11-07
add undefinedFormatting checker

peakle:


peakle:
tell me if pr look good, i will create pr to extern-testdata

PR #1159, opened 2021-11-12
add protoGetter checker

ghostiam:


cristaloleg:
Hi, can you add a description what is the purpose of this check? Thank you!

ghostiam:
@cristaloleg If there are empty structures in the received request or grpc response, then when receiving fields from this structure, we will get panic, but if we use the getter, we will get ZeroValue.

Getters look like this:
```go
func (x *Embedded) GetS() string {
	if x != nil {
		return x.S
	}
	return ""
}

func (x *Embedded) GetEmbedded() *Embedded {
	if x != nil {
		return x.Embedded
	}
	return nil
}
```

ghostiam:
@cristaloleg please, rerun build. I fixed linter warnings.

PR #1169, opened 2021-12-23
Add checker for TODO comments without detail/assignee

### LINKED ISSUE #434: lint: TODO/FIX(ME)/BUG/XXX comments without addressing a person
Before:
```go
// TODO
fiiWithCtx(nil, a, b)
```
After:
```go
// TODO(admin): pass context.TODO() instead of nil
fiiWithCtx(nil, a, b)
```

liamg:
Resolves #434


cristaloleg (approved):
sgtm

quasilyte:
I think it should be marked with an "opinionated" tag, just in case.
It's not always sensible to require this rule to be followed.
We can merge PR and add this tag separately though, as it's marked with experimental anyway.

checkers/todoCommentWithoutDetail_checker.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+type todoCommentWithoutCodeChecker struct {
+	astwalk.WalkHandler
+	ctx   *linter.CheckerContext
+	regex *regexp.Regexp
+}
+
+func (c *todoCommentWithoutCodeChecker) VisitComment(cg *ast.CommentGroup) {
+	comment := strings.TrimSpace(cg.Text())
```
quasilyte, line 41: Using `Text()` method is generally slower than using `cg.List[...].Text` fields directly.
I think we can analyze it in line-by-line fashion and avoid TrimSpace call as well since regexp already includes a space skipping part.
What do you think?

liamg: @quasilyte Thanks, I've addressed this. Sorry for the delay. It looks like using the `cg.List[...].Text` fields include the comment initialisers (`//` and `/*`) so I've adjusted the regex to take this into account and everything seems to work.

liamg:
@quasilyte Thanks, I've addressed your comment. Sorry for the delay. It looks like the `cg.List[...].Text` fields include the comment initialisers (`//` and `/*`) so I've adjusted the regex to take this into account and everything seems to work.

quasilyte:
I have a few more suggestions, but I'm OK with merging it as is, so they can be addressed later.
One of my questions is: this is probably more style-related than correctness-related. So the "diagnostic" tag is quite misleading.

liamg:
Cool, I changed it to `style` and added `opinionated` too.

quasilyte (approved):

PR #1175, opened 2021-12-26
made go-critic's rules exportable

peakle:


cristaloleg:
But why?

checkers/rules/rules.go
```diff
@@ -4,6 +4,9 @@ import (
 	"github.com/quasilyte/go-ruleguard/dsl"
 )
 
+// Bundle - used to import rules by external packages
+var Bundle = dsl.Bundle{}
```
quasilyte, line 8: Unfortunately, we're using rules embedding in go-critic for builtin rules.
This means we can't export rules normally:
https://github.com/quasilyte/go-ruleguard/issues/355

This is not a hard issue to fix in ruleguard, but right now it's not an option to use bundle vars in linters that use a precompile step.

quasilyte:
We really need docker (or whatever) tests to check that gocritic can work without installed dsl dependency for the cases where custom ruleguard rules are not needed.

Embedded rules should not require dsl package.

Issue #1177, opened 2021-12-28
Move code related to embedded rules from checkers/checkers.go to checkers/embedded_rules.go


quasilyte:
This includes `init()` and `go:generate` comment plus everything else.

cristaloleg:
Fixed #1186

PR #1186, opened 2022-01-02
Move out embedded rules from checkers/checkers.go

JuozasVainauskas:
Move out embedded rules from checkers/checkers.go to checkers/embedded_rules.go

Issue: https://github.com/go-critic/go-critic/issues/1177

### DIFF
--- a/checkers/checkers.go
+++ b/checkers/checkers.go
@@ -2,15 +2,8 @@
 package checkers
 
 import (
-	"fmt"
-	"go/ast"
-	"go/build"
-	"go/token"
-	"os"
-
-	"github.com/go-critic/go-critic/checkers/rulesdata"
 	"github.com/go-critic/go-critic/framework/linter"
-	"github.com/quasilyte/go-ruleguard/ruleguard"
+	"os"
 )
 
 var collection = &linter.CheckerCollection{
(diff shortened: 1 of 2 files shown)


checkers/checkers.go
```diff
@@ -2,15 +2,8 @@
...
-	"fmt"
-	"go/ast"
-	"go/build"
-	"go/token"
-	"os"
-
-	"github.com/go-critic/go-critic/checkers/rulesdata"
 	"github.com/go-critic/go-critic/framework/linter"
-	"github.com/quasilyte/go-ruleguard/ruleguard"
+	"os"
```
cristaloleg, line 6: Nit: it would be better to move"os" before ""github.com/go-critic/go-critic/framework/linter" and keep new line between them ;)

checkers/embedded_rules.go
```diff
@@ -0,0 +1,102 @@
...
+
+import (
+	"fmt"
+	"github.com/go-critic/go-critic/checkers/rulesdata"
+	"github.com/go-critic/go-critic/framework/linter"
+	"github.com/quasilyte/go-ruleguard/ruleguard"
+	"go/ast"
+	"go/build"
+	"go/token"
+	"os"
```
cristaloleg, line 11: Nit: also here, stdlib goes before all other deps and newline between groups ;)

cristaloleg (approved):
SGTM 🚀

cristaloleg (approved):

quasilyte:
Nice. I was always struggling to remember where do we keep the code related to the embedded rules.

quasilyte (approved):

Issue #1183, opened 2022-01-01
Investigate the performance regression reason


quasilyte:
https://github.com/golangci/golangci-lint/pull/2041#issuecomment-956871099

PR #1188, opened 2022-01-06
optimize deprecatedComment_checker: delete regexp usage

peakle:
pr for: https://github.com/go-critic/go-critic/issues/1183
benchmarks in progress

### DIFF
--- a/checkers/deprecatedComment_checker.go
+++ b/checkers/deprecatedComment_checker.go
@@ -116,7 +116,11 @@ func (c *deprecatedCommentChecker) VisitDocComment(doc *ast.CommentGroup) {
 
 		// Check for other commonly used patterns.
 		for _, pat := range c.commonPatterns {
-			if strings.HasPrefix(strings.ToLower(l), pat) {
+			if len(l) < len(pat) {
+				continue
+			}
+
+			if strings.EqualFold(l[:len(pat)], pat) {
 				c.warnPattern(comment)
 				return
 			}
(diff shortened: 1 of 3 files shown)


checkers/deprecatedComment_checker.go
```diff
@@ -114,7 +116,7 @@ func (c *deprecatedCommentChecker) VisitDocComment(doc *ast.CommentGroup) {
 
 		// Check for other commonly used patterns.
 		for _, pat := range c.commonPatterns {
-			if pat.MatchString(l) {
+			if strings.HasPrefix(strings.ToLower(l), pat) {
```
quasilyte, line 119: It's possible to use EqualFold here, if you take the prefix of `l` of `len(pat)`.
This will avoid allocations in here.

quasilyte:
We may need more tests to make sure that nothing is broken after this optimization. Could you add some, just to be sure?

peakle:
> We may need more tests to make sure that nothing is broken after this optimization. Could you add some, just to be sure?

added test cases

quasilyte:
Benchmarks are not necessary if you can measure that gocritic becomes faster after these changes.

peakle:
> Benchmarks are not necessary if you can measure that gocritic becomes faster after these changes. The code looks good so far.
> 
> Maybe https://github.com/sharkdp/hyperfine can help here.

i executed old and new version on githube.com/kubernetes/kubernetes and as i see call graph looks better

but also i found that now it impossible to parse multiline comments like this one:
```
// ComponentStatusList represents the list of component statuses
// Deprecated: This API is deprecated in v1.19+
type ComponentStatusList struct {}
```

PR #1191, opened 2022-01-07
checkers: add fixable for commentFormatting (#1190)

### LINKED ISSUE #1190: Support autofix for commentFormatting

okhowang:
Fixes #1190

### DIFF
--- a/checkers/commentFormatting_checker.go
+++ b/checkers/commentFormatting_checker.go
@@ -77,5 +77,9 @@ func (c *commentFormattingChecker) specialChar(r rune) bool {
 }
 
 func (c *commentFormattingChecker) warn(comment *ast.Comment) {
-	c.ctx.Warn(comment, "put a space between `//` and comment text")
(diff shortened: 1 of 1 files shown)


cristaloleg (approved):
sgtm

quasilyte:
This will add extra spaces inside comments:
```
//Comment with //inside other comment
=>
// Comment with // inside other comment
```
I don't think that it's a big deal, but it makes it less reliable in general as it does more changes than it should.

okhowang:
> This will add extra spaces inside comments:
> 
> ```
> //Comment with //inside other comment
> =>
> // Comment with // inside other comment
> ```
> 
> I don't think that it's a big deal, but it makes it less reliable in general as it does more changes than it should.

I use `strings.Replace`  with n=1 which mean replace only first `//`

quasilyte:
Ah, that's true.

quasilyte (approved):

Issue #1167, opened 2021-12-17
A new performance check: prefer operator comparisons to strings.Compare


go101:
The current implementation of the `strings.Compare` function is not efficient, in a some deliberate way.
refs:
* https://github.com/golang/go/issues/50167
* https://github.com/golang/go/issues/50209

At least, for non-three-way comparisons, using operators to compare strings is always faster than using the `strings.Compare` function.

peakle:
@go101 
you can try to open pr to https://github.com/quasilyte/go-perfguard with optimization rule like:
```m.Match(`strings.Compare($_, $_)`).Report(`strings.Compare in hot path`)```


contribution guide: https://github.com/quasilyte/go-perfguard/blob/master/_docs/CONTRIBUTING.md

quasilyte:
I would add them to go-critic though.
go-critic is something that is more useful for most users.
perfguard is an experiment that has ~0 integrations.

quasilyte:
By the way.

@go101 thank you for your books. :heart:

peakle:
@quasilyte can be closed as i see

quasilyte:
I think that we haven't implemented all of the suggested patterns.
Some of them can't be expressed in ruleguard rules beautifully, so maybe someone would like to create a separate checker for that.

PR #1195, opened 2022-01-16
checkers: add new checker stringsCompare

peakle:
pr for: https://github.com/go-critic/go-critic/issues/1167


checkers/rules/rules.go
```diff
@@ -736,3 +736,34 @@ func dynamicFmtString(m dsl.Matcher) {
...
 		Report(`use errors.New($f($*args)) or fmt.Errorf("%s", $f($*args)) instead`)
 }
+
+//doc:summary Detects strings.Compare usage
+//doc:tags    performance experimental
+//doc:before  strings.Compare(x, y)
+//doc:after   x < y
+func stringsCompare(m dsl.Matcher) {
+	m.Match(`strings.Compare($s1, $s2) == 0`).
+		Report(`don't use strings.Compare on hot path, change it to built-in operators`).
```
quasilyte, line 746: This warning message is not helpful.
Suggest template is more explanatory.
In general, the warning message should tell how to fix the situation.
In other words, it should tell what to do instead of saying what not to do.

checkers/rules/rules.go
```diff
@@ -736,3 +736,34 @@ func dynamicFmtString(m dsl.Matcher) {
 		Suggest("errors.New($f($*args))").
 		Report(`use errors.New($f($*args)) or fmt.Errorf("%s", $f($*args)) instead`)
 }
+
+//doc:summary Detects strings.Compare usage
+//doc:tags    performance experimental
```
quasilyte, line 741: It's up to a debate whether this is a performance-first issue.
I would say most patterns would fall to style category, like using `==` instead of `strings.Compare`.
Other, less obvious patterns, can go to a separate, performance-tagged checker.

checkers/rules/rules.go
```diff
@@ -736,3 +736,34 @@ func dynamicFmtString(m dsl.Matcher) {
...
+		Suggest(`$s1 < $s2`).
+		At(m["s1"])
+
+	m.Match(`strings.Compare($s1, $s2) > 0`,
+		`strings.Compare($s1, $s2) == 1`).
+		Report(`don't use strings.Compare on hot path, change it to built-in operators`).
+		Suggest(`$s1 > $s2`).
+		At(m["s1"])
+
+	m.Match(`$_ = strings.Compare($s1, $_)`,
```
quasilyte, line 762: This is too general and will mark any usages of `strings.Compare`.
Giving false positives is bad.

quasilyte:
I would do this in 2 steps:
1. suggest simpler operations for the cases where the replacement is obvious.
2. Implement other cases as a hand-written analyzer, to avoid false positives.

quasilyte:
@cristaloleg, what do you think?

checkers/rules/rules.go
quasilyte: Maybe this message comes from go-perfguard style warnings.
But there is a difference there: it doesn't report anything if it has a replacement, so it just reports the suggestion template there. If there is no suggestion available, it can only say "don't do that" because it doesn't know how to fix the code.

In my opinion, reporting performance issue without a fix is only useful if it's a big performance issue.

checkers/rules/rules.go
peakle: +

checkers/rules/rules.go
peakle: changed to style

checkers/rules/rules.go
peakle: \+

checkers/rules/rules.go
```diff
@@ -736,3 +736,24 @@ func dynamicFmtString(m dsl.Matcher) {
 		Suggest("errors.New($f($*args))").
 		Report(`use errors.New($f($*args)) or fmt.Errorf("%s", $f($*args)) instead`)
 }
+
+//doc:summary Detects strings.Compare usage
+//doc:tags    style experimental
+//doc:before  strings.Compare(x, y)
+//doc:after   x < y
+func stringsCompare(m dsl.Matcher) {
+	m.Match(`strings.Compare($s1, $s2) == 0`, `0 == strings.Compare($s1, $s2)`).
```
quasilyte, line 745: There is no need to add a yoda-style pattern here.

They should be reported by a separate checker that just checks for `0 == $_` (not exactly, but that's the idea).

peakle: \+

quasilyte (changes requested):
Please, cover all cases with tests.
I don't see -1 value case.
Whether possible, all corner cases should be tested.

It's also a good practice to have "fixed" code examples in the negative tests suite.
For example, if linter suggests `>` operator, negative tests suite should contain the suggested code snippet, to be sure that this replacements does not trigger any other warning.

quasilyte (approved):

quasilyte:
LGTM.
@cristaloleg wanna make a final review? :)

quasilyte (approved):

PR #1196, opened 2022-01-21
checkers/missingErrorAccess_checker: new checker

komkom:
# MissingErrAccess Checker

When a function returns an error the checker attempts to figure out if one of returned none error variables gets accessed before the actual error. 
This checker should only issue a warning if such an access actually happened. As a consequence, there are also cases where this access stays undetected.

Many cases are documented in this file
```
checkers/testdata/missingErrAccess/all_tests.Go
```


checkers/testdata/missingErrAccess/all_tests.go
```diff
@@ -0,0 +1,262 @@
...
+	/*! defer, missing error check accessing [f] */
+	defer f()
+	if err != nil {
+
+	}
+
+	a, err := recv1()
+	msg := err.Error() + strconv.Itoa(a)
+	_ = msg
+}
```
komkom, line 262: Why are the tests below falling for this PR?

komkom: Its a bit said that nobody is reviewing this ... Is this repo dead?

PR #1205, opened 2022-02-15
ruleguard: add filter by tags for analyzer mode

peakle:
tested in: https://github.com/delivery-club/delivery-club-rules/pull/42


quasilyte:
Are you sure that the default ruleset is unchanged?
Tests are failing for some reason.

peakle:
> Are you sure that the default ruleset is unchanged? Tests are failing for some reason.

i changed precompiled rules according to "precompiled check" on ci, will try investigate fails reason on testdata with debug-* flag tonight

peakle:
> Are you sure that the default ruleset is unchanged? Tests are failing for some reason.

as i see it because you add in https://github.com/quasilyte/go-ruleguard/pull/353 truncate long messages, so some test failed because part of full message was replaced by `<...>` pattern

peakle:
testdata update: https://github.com/go-critic/extern-testdata/pull/4

checkers/ruleguard_checker.go
```diff
@@ -135,20 +135,41 @@ func newRuleguardChecker(info *linter.CheckerInfo, ctx *linter.CheckerContext) (
...
+		}
+
 		disabledGroups[g] = true
 	}
 	flagEnable := info.Params.String("enable")
 	if flagEnable != "<all>" {
 		for _, g := range strings.Split(flagEnable, ",") {
 			g = strings.TrimSpace(g)
+			if t := strings.Split(g, "#"); len(t) == 2 {
+				enabledTags[t[1]] = true
```
quasilyte, line 155: How about?
```
if strings.HasPrefix(g, "#") {
  enabledTags[strings.TrimPrefix(g, "#")] = true
}
```
It looks like you're splitting a string like `#tag` just to check if it will result in 2 elements, making `foo#tag` a valid input.

quasilyte:
Looks OK, but I'm not sure about that strings.Split usage.
Also, there are integration tests (`tests/_integration`) that allow testing gocritic with arbitrary flags and inputs.
You can test the new flags there.

checkers/ruleguard_checker.go
peakle: fixed

quasilyte (approved):

PR #1207, opened 2022-02-16
all: drop bright package

### LINKED ISSUE #1206: Update logrusorgru version
Faced with the problem of using go-critic in the company because of the WTFPL license logrusorgru/aurora. Starting with version 2.0.1 they updated the license.
Could you please update the version of this library?

cristaloleg:
Fixes #1206


quasilyte:
Maybe we need to remove the colored output cli option too?

quasilyte (approved):

PR #1211, opened 2022-03-24
Remove the `experimental` tag from the `ioutil` check so that it's enabled by default

jeffwidman:
The `ioutil` package was [deprecated as part of `go` `1.16`](https://go.dev/doc/go1.16#ioutil).

Now that `go` `1.18` has dropped, the core `go` team no longer supports
`1.16`.

So it should be safe to remove the `experimental` tag. This will switch
the check from disabled by default to enabled by default.

Note that the check still has internal `if` statements to only fire on code
running `1.16` or newer, so even though it's enabled by default, it won't
start complaining on code running older versions of `go`.

### DIFF
--- a/checkers/rules/rules.go
+++ b/checkers/rules/rules.go
@@ -40,7 +40,7 @@ func deferUnlambda(m dsl.Matcher) {
 }
 
 //doc:summary Detects deprecated io/ioutil package usages
-//doc:tags    style experimental
+//doc:tags    style
 //doc:before  ioutil.ReadAll(r)
 //doc:after   io.ReadAll(r)
 func ioutilDeprecated(m dsl.Matcher) {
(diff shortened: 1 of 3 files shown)


jeffwidman:
I also opened https://github.com/golang/go/issues/51927 to see about marking `ioutil` as deprecated in `go` itself, as then this check could be completely removed.

jeffwidman:
Looks like upstream may mark them as `"Deprecated: "` in `1.19`, so rather than making experimental, could simply remove this check altogether once `1.19` lands.

quasilyte (approved):

quasilyte:
Could you re-run `go generate ./...` after rebasing on gocritic master?

Pseudo algorithm:

```
git checkout master
git pull origin master
git checkout yourbranch
git rebase master
go generate ./...
git add -u
git commit --amend
git push -f origin yourbranch
```

jeffwidman:
Thanks, but I'm actually going to close this instead in favor of simply removing the check now that my patch to `go 1.19` looks likely to land. See my rationale in https://github.com/go-critic/go-critic/pull/1212#issuecomment-1080897377.

This PR simply reinforces the point to me that there's a maintenance cost to keeping this check around.

PR #1212, opened 2022-03-25
checkers: remove the `ioutil` check

jeffwidman:
In light of https://github.com/golang/go/issues/51927, `ioutil` will get
the `"Deprecated; "` annotation starting in `go` `1.19`.

As a result, this custom checker will be superfluous because linters
that mark any deprecated code usage will also complain about `ioutil`.

So removing this check.


jeffwidman:
This PR should probably just float til `go` `1.19` lands...

quasilyte:
Well, another way is to add a second filter constraint to the go version, so it's a range:
```
m.GoVersion().GreaterEqThan("1.16") && m.GoVersion().LessEqThan("1.19")
```

jeffwidman:
> Well, another way is to add a second filter constraint to the go version, so it becomes a range:

Yeah, I considered that, but IMO it's not worth paying the maintenance tax over time given that the `go` community is generally pretty quick to upgrade to newer releases. Especially any teams running `go-critic`, they'd be even more likely to work to stay on new releases.

Given that the `ioutil` code isn't actually going to be removed until Go 2, and it's an easy migration to flip from old function locations to new function locations, so it's not that bad if the usage slowly creeps into a codebase for a few more months because it's not breaking anything and the migration will be easy... so I don't see the value in keeping this check around once `go 1.19` drops.

jeffwidman:
I rebased and re-ran `go generate ./...`, thanks for pointing out the need for that in https://github.com/go-critic/go-critic/pull/1211#issuecomment-1080463837.

checkers/testdata/octalLiteral/negative_tests.go
```diff
@@ -66,7 +65,7 @@ func NoWarningsFs() {
 	_ = fs.FileMode(0o555)
 }
 
-func NoWarningsOs() {
+func NoWarningsOsOpenFile() {
```
jeffwidman, line 68: Alternatively, I could move the `os.OpenFile()` and `os.WriteFile()` calls into a single `NoWarningsOs()`. Flipped a coin and went with the more explicit choice...

checkers/testdata/octalLiteral/positive_tests.go
```diff
@@ -76,7 +75,12 @@ func (os *OpenServer) Init(x int) {
 	os.x = x
 }
 
-func warningsOs() {
+func warningsFs() {
+	/*! use new octal literal style, 0o555 */
+	_ = fs.FileMode(0555)
+}
```
jeffwidman, line 81: Nothing changed with this func, I simply re-ordered it above the `warningsOs*` funcs for alphabetical order/consistency with `negative_tests.go`.

checkers/testdata/octalLiteral/positive_tests.go
```diff
@@ -76,7 +75,12 @@ func (os *OpenServer) Init(x int) {
...
+
+func warningsOsOpenFile() {
```
jeffwidman, line 83: Similar to `negative_tests.go`, this could be a single `warningsOs()` including both the `OpenFile()` and the `WriteFile()` calls.

docs/overview.md
```diff
@@ -2802,5 +2778,3 @@ return nil != ptr
 ```go
 return ptr != nil
 ```
-
-
```
jeffwidman, line 2806: I noticed extraneous whitespace in a few places... does your CI not flag that?

quasilyte:
AFAIK, go1.18 is just being released.
There are at least a few months before 1.19 and maybe + a couple of months before most people would switch to that newer version. Are you sure we want to remove a checker that will be useful in the next ~half of the year?

jeffwidman:
That's up to you. 

I figured you'd just let this PR float until go 1.19 actually lands, but totally up to you. 

I honestly doubt that many folks use it since it's got the `experimental` tag, so not enabled by default...

jeffwidman:
`go` `1.19` just dropped, so this PR is now ready for review...

jeffwidman:
Should I close this or fix the merge conflicts?

jeffwidman:
nudge...

cristaloleg:
Hey @jeffwidman , sorry that it took so long, can you rebase, pleasse?

cristaloleg:
Thanks!

Issue #1223, opened 2022-05-25
rangeValCopy panic on generic array ranges


gabizou:
Some perfectly compilable Go 1.18 code:
```go
func FilterSlice[T any](original []T, f func(T) bool) []T {
   var filtered []T
   for _, o := range original {
      if f(o) {
         filtered = append(filtered, o)
      }
   }
   return filtered
}
```
The [rangeValChecker](https://github.com/go-critic/go-critic/blob/bf8c01ca6692a5a3e8c20b1caef47dfab1412c88/checkers/rangeValCopy_checker.go#L68), calling [go/types/sizes.SizeOf, which for generics has an assertion not to do](https://github.com/golang/go/blob/78b722d8c2f764c3048c6f0344e9ebcd2687813d/src/go/types/sizes.go#L156-L159) 

While this code will not panic:
```go
func FilterSlice[T any](original []T, f func(T) bool) []T {
   var filtered []T
   for i :=0; i < len(original); i++ {
      o := original[i]
      if f(o) {
         filtered = append(filtered, o)
      }
   }
   return filtered
}
```
Note: I didn't run this specifically via go-critic, but via golangci-lint, so the stack trace is cluttered with golangci-lint's code, but still checked that it is using `go-critic` v0.6.3

<summary>

```
$ golangci-lint run
ERRO [runner] Panic: gocritic: package "service" (isInitialPkg: true, needAnalyzeSource: true): assertion failed: goroutine 11478 [running]:
runtime/debug.Stack()
        runtime/debug/stack.go:24 +0x68
github.com/golangci/golangci-lint/pkg/golinters/goanalysis.(*action).analyzeSafe.func1()
        github.com/golangci/golangci-lint/pkg/golinters/goanalysis/runner_action.go:101 +0x104
panic({0x1032a0780, 0x10340d5a0})
        runtime/panic.go:838 +0x204
go/types.assert(...)
        go/types/errors.go:21
go/types.(*StdSizes).Sizeof(0x140001152d0, {0x103414f88?, 0x14005c18b70})
        go/types/sizes.go:159 +0x3a0
github.com/go-critic/go-critic/checkers.(*rangeValCopyChecker).VisitStmt(0x14004118940, {0x103417e70?, 0x140037c8e60?})
        github.com/go-critic/go-critic@v0.6.3/checkers/rangeValCopy_checker.go:68 +0xb0
[...]

PR #1237, opened 2022-07-08
don't call SizeOf for type parameters

mcdoker18:
StdSizes.SizeOf method doesn't support type parameters. https://github.com/golang/go/blob/78b722d8c2f764c3048c6f0344e9ebcd2687813d/src/go/types/sizes.go#L156-L159

Fixes go-critic/go-critic#1223

### DIFF
--- a/checkers/hugeParam_checker.go
+++ b/checkers/hugeParam_checker.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/go-critic/go-critic/checkers/internal/astwalk"
 	"github.com/go-critic/go-critic/framework/linter"
+	"golang.org/x/exp/typeparams"
 )
 
 func init() {
@@ -49,6 +50,9 @@ func (c *hugeParamChecker) checkParams(params []*ast.Field) {
 	for _, p := range params {
 		for _, id := range p.Names {
 			typ := c.ctx.TypeOf(id)
+			if _, ok := typ.(*typeparams.TypeParam); ok {
+				continue
+			}
 			size := c.ctx.SizesInfo.Sizeof(typ)
 			if size >= c.sizeThreshold {
 				c.warn(id, size)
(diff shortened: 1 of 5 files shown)


cristaloleg (dismissed):
SGTM

mcdoker18:
```
hugeParam_checker.go:53:28: TypeParam not declared by package types
```

@cristaloleg  I use go1.18 locally, so I didn't mention that go-critic still uses 1.17.

Do you have any plan\issue to upgrade go-critic to 1.18? I can try in separate pull request

cristaloleg:
Well, that's a hard question to be honest. We should bump go-critic to Go 1.18 somewhere in July. CC: @quasilyte

checkers/hugeParam_checker.go
```diff
@@ -49,6 +50,9 @@ func (c *hugeParamChecker) checkParams(params []*ast.Field) {
 	for _, p := range params {
 		for _, id := range p.Names {
 			typ := c.ctx.TypeOf(id)
+			if _, ok := typ.(*types.TypeParam); ok {
```
quasilyte, line 53: Shouldn't we use a polyfill like `x/exp/typeparams` to make this code compileable with older Go versions?
golangci-lint can't deprecate pre-generics release yet AFAIK.

quasilyte: https://pkg.go.dev/golang.org/x/exp/typeparams

mcdoker18: Fixed.

If we support go 1.17 then I should hide the new test cases using build constraints.

mcdoker18: BTW, If you'd like I can add go1.18 CI support in a separated PR.

https://github.com/mcdoker18/go-critic/pull/3/commits/2041b982633b256b73c61521614172b11133f1dd

quasilyte:
The code looks good now.

mcdoker18:
@cristaloleg rebased with master

quasilyte (approved):

mcdoker18:
@cristaloleg Hi, can you take a look?

PR #1254, opened 2022-08-25
docs: update README install instruction for go 1.18

Faagerholm:
As of go 1.18 `go get` outside of a module is deprecated, we should use `go install` instead. 
This PR adds a small note on this for new users who wish to download go-critic


cristaloleg (approved):
Thanks!

cristaloleg (approved):
Thanks!

PR #1260, opened 2022-08-28
checkers: fix reflect.Value suggestion in redundantSprint

### LINKED ISSUE #1255: redundantSprint false positive with reflect.Value
https://pkg.go.dev/reflect#Value.String
> The fmt package treats Values specially. It does not call their String method implicitly but instead prints the concrete values they hold.

So `fmt.Sprint(v)` is not equivalent to `v.String()` (if `v` is a `reflect.Value`).

https://github.com/go-critic/go-critic/blob/5c89720acc9e12fbbc23d95b381ee3b594762088/checkers/rules/rules.go#L11-L26
Is it possible to exclude `reflect.Value` ?

quasilyte:
Also enable embedded rules testing: we need to call
InitEmbeddedRules() somewhere inside the testing code
in order for them to be registered.

Fixes #1255

### DIFF
--- a/checkers/checkers_test.go
+++ b/checkers/checkers_test.go
@@ -12,6 +12,12 @@ import (
 	"github.com/google/go-cmp/cmp"
 )
 
+func init() {
+	if err := InitEmbeddedRules(); err != nil {
+		panic(err) // Should never happen
+	}
+}
+
 func TestCheckers(t *testing.T) {
 	allParams := map[string]map[string]interface{}{
 		"captLocal": {"paramsOnly": false},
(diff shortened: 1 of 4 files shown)


checkers/checkers_test.go
```diff
@@ -12,6 +12,12 @@ import (
 	"github.com/google/go-cmp/cmp"
 )
 
+func init() {
+	if err := InitEmbeddedRules(); err != nil {
```
cristaloleg, line 16: Is there a sense to keep this func exported after this PR ?

I'm not proposing to drop it, more like deprecated and remove in the future. WDYT?

cristaloleg (approved):

checkers/checkers_test.go
quasilyte: I believe golangci-lint may still want to avoid registering the embedded rules if gocritic is disabled.

quasilyte:
I fixed the external-tests code (removed the false positive reports).
Let me run the tests again.

cristaloleg (approved):

PR #1273, opened 2022-10-05
fix(newDeref): ignore generics

### LINKED ISSUE #1272: newDeref: possible false-positive with generics
First of all, thank you for the awesome linter, I'm in love with go-critic ❤️

It seems like [newDeref](https://go-critic.com/overview.html#newderef) reports a false-positive when type parameters are used:

```go
func test[T any]() T {
	return *new(T)
}
```

> newDeref: replace `*new(T)` with `T(nil)`

Which is not possible because the conversion is invalid:

> cannot convert nil to T ​compiler: InvalidConversion

Found [here](https://github.com/junk1tm/assert/blob/main/assert.go#L105).

golangci-lint version: 1.49.0
gocritic version: v0.6.5 (installed with brew)

junk1tm:
fixes #1272

### DIFF
--- a/checkers/newDeref_checker.go
+++ b/checkers/newDeref_checker.go
@@ -7,6 +7,7 @@ import (
 	"github.com/go-critic/go-critic/checkers/internal/lintutil"
 	"github.com/go-critic/go-critic/framework/linter"
 	"github.com/go-toolsmith/astcast"
+	"golang.org/x/exp/typeparams"
 	"golang.org/x/tools/go/ast/astutil"
 )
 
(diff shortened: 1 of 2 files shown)


junk1tm:
@quasilyte @cristaloleg would you take a look? thank you

junk1tm:
Also I've noticed that the `badRegexp` test does not pass with go 1.19.1 🤔

> make test-checker
> --- FAIL: TestCheckers/badRegexp (0.05s)

Is it ok?

quasilyte (approved):

quasilyte:
> Is it ok?

You can open a new issue about it. I haven't checked 1.19 yet.

PR #1276, opened 2022-10-08
load rules once

peakle:

PR #1280, opened 2022-10-11
add new method for ruleguard warnings

peakle:
add new warn method without ast.Node usage: https://github.com/go-critic/go-critic/pull/1276


quasilyte (approved):
LGTM.

PR #1281, opened 2022-10-13
checkers: removed timeCmpSimplify checker

### LINKED ISSUE #1257: timeCmpSimplify rule is wrong
Rule is wrong because times on the condition can be equal and in this case, not after isn't same with before.

```go
package main

import (
	"fmt"
	"time"
)

func main() {
	t := time.Now()
	fmt.Println(!t.After(t))
	fmt.Println(t.Before(t))
}
```

https://go.dev/play/p/dAY9FXrxzOA

quasilyte:
The checker suggested a change that is not equivalent to the code being replaced. The results were different when the time values were identical.

This issue has no obvious fix, so the checker is removed completely. It was marked as experimental after all.

Fixes #1257


cristaloleg (approved):
Sad but true.

jufemaiz:
LGTM.

jufemaiz:
A year on should this get a move on and merged in?

cristaloleg:
Indeed, I will merge this today, thanks for the reminder.

PR #1289, opened 2022-11-08
checkers/wrapperFunc: suggest `strings.Cut` instead of `strings.Index`

junk1tm:
This change utilize `strings.Cut`, a new function added in go 1.18, which turned out [quite useful](https://go-review.googlesource.com/c/go/+/351711).

Unfortunately, the current implementation does not cover all cases, e.g. the same examples won't match if the `:=` operator is used instead of `=`. I wonder if we should just copy-paste the same checks with different operators 🤔

Also, a similar function exists in the `bytes` package, so we could implement the `bytes.Index` -> `bytes.Cut` check as well (in another PR, I guess).


cristaloleg (approved):
SGTM

cristaloleg:
Yo, can you rebase your changes, please?

junk1tm:
Sure, will do 👍

cristaloleg:
Merged to temp branch, thanks.

PR #1290, opened 2022-11-08
checkers/wrapperFunc: suggest `strings.Contains*` instead of `strings.Index* >= 0`

junk1tm:
This change utilize the fact that the `strings.Contains*` functions are just wrappers for `strings.Index* >= 0`. The same goes for the `bytes` package.


cristaloleg (approved):
SGTM

cristaloleg (approved):

cristaloleg:
Thanks! and sorry for a delay.

PR #1316, opened 2023-02-20
deps: bump

cristaloleg:


dependabot.yml
```diff
@@ -1,10 +0,0 @@
-version: 2
```
cristaloleg, line 1: Not needed, we've same file in `.github`

quasilyte (approved):

PR #1322, opened 2023-02-23
checkers: add sloppyTestFuncName

cristaloleg:


checkers/rules/rules.go
```diff
@@ -785,3 +785,27 @@ func uncheckedInlineErr(m dsl.Matcher) {
...
+			!m["test"].Text.Matches("test.*")).
+		Report("function $test should be of form TestXXX(t *testing.T)")
+
+	m.Match(`func $bench($_ *testing.B) { $*_ }`).
+		Where(!m["bench"].Text.Matches("Benchmark.*") &&
+			!m["bench"].Text.Matches("bench.*")).
+		Report("function $bench should be of form BenchmarkXXX(b *testing.B)")
+
+	m.Match(`func $test($_ *testing.T) { $*_ }`).
+		Where(m["test"].Text.Matches("test.*")).
```
quasilyte, line 805: Doesn't it need anchors like `^test.*$`?
Because otherwise I think it will match things like `paratestBlah`.

cristaloleg: Thanks, fixed!

quasilyte (approved):

PR #1366, opened 2023-09-04
checker: add badSyncOnceFunc

cristaloleg:
Got bitten today.

Covers 2 cases:
- when `sync.OnceFunc` result is invoked in-place (can be called many times)
- when `sync.OnceFunc` result isn't assigned to a variable (effectively no-op)


checkers/rules/rules.go
```diff
@@ -808,3 +808,17 @@ func sloppyTestFuncName(m dsl.Matcher) {
...
 		Report("function $bench looks like a benchmark helper, consider to change 1st param to 'tb testing.TB'")
 }
+
+//doc:summary Detects bad usage of sync.OnceFunc
+//doc:tags    diagnostic experimental
+//doc:before  sync.OnceFunc(foo)()
+//doc:after   fooOnce := sync.OnceFunc(foo); ...; fooOnce()
+func badSyncOnceFunc(m dsl.Matcher) {
+	m.Match(`$*_; sync.OnceFunc($x); $*_;`).
+		Suggest("sync.OnceFunc($x)()").
```
cristaloleg, line 818: Here I'm proposing to just invoke the result and after that match below will say "hey, it's still wrong".

checkers/rules/rules.go
```diff
@@ -808,3 +808,17 @@ func sloppyTestFuncName(m dsl.Matcher) {
...
+		Where(m.GoVersion().GreaterEqThan("1.21"))
+
+	m.Match(`sync.OnceFunc($x)()`).
+		Report("possible sync.OnceFunc misuse, consider to assign sync.OnceFunc($x) to a variable").
```
cristaloleg, line 822: Probably this message can be better (and shorter). Better ideas?

quasilyte (approved):

PR #1372, opened 2023-09-23
checkers/hugeParam: ignore method String() string.

### LINKED ISSUE #1168: Consider ignoring some common methods in hugeParam checker
For example, `String() string` method sometimes is defined on the value to make it possible to pass it as `%s` value.

```
package main

import "fmt"

type examplePtr struct{}

func (*examplePtr) String() string { return "PTR" }

type exampleValue struct{}

func (exampleValue) String() string { return "VALUE" }

func main() {
	ptr1 := examplePtr{}
	ptr2 := &examplePtr{}
	val1 := exampleValue{}
	val2 := &exampleValue{}
	fmt.Printf("%s %s %s %s\n", ptr1, ptr2, val1, val2)
}
```
https://go.dev/play/p/s7MxcI4YVWX
For ptr receiver, `ptr1` will not produce `PTR` string.

Lasiar:
Fixes #1168.

### DIFF
--- a/checkers/hugeParam_checker.go
+++ b/checkers/hugeParam_checker.go
@@ -5,6 +5,8 @@ import (
 
 	"github.com/go-critic/go-critic/checkers/internal/astwalk"
 	"github.com/go-critic/go-critic/linter"
+
+	"github.com/go-toolsmith/astcast"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


cristaloleg (approved):
Nice, thanks!

quasilyte:
You folks are too fast. I can't even manage to read a notification in 1 hour while you go a whole road in this period of time. :D

PR #1393, opened 2024-01-01
checkers/emptyStringTest: add len(s) >= 0 case

### LINKED ISSUE #1392: emptyStringTest: handle len(s) > 0
sample code
```go
package main

import "fmt"

func main() {
        s := ""
        if len(s) == 0 { fmt.Println("len(s) == 0") }
        if len(s) != 0 { fmt.Println("len(s) != 0") }
        if len(s) > 0 { fmt.Println("len(s) > 0") }
}
```

`gocritic check -enableAll main.go` gives only
```
./main.go:8:5: emptyStringTest: replace `len(s) != 0` with `s != ""`
./main.go:7:5: emptyStringTest: replace `len(s) == 0` with `s == ""`
```

and no complaints to replace `len(s) > 0` with `s != ""`

cristaloleg:
Fixes #1392


checkers/rules/rules.go
```diff
@@ -160,10 +160,22 @@ func emptyStringTest(m dsl.Matcher) {
...
+		Where(m["s"].Type.Is(`string`)).
+		Report("replace `$$` with `$s != \"\"`")
 
 	m.Match(`len($s) == 0`).
 		Where(m["s"].Type.Is(`string`)).
 		Report("replace `$$` with `$s == \"\"`")
+	m.Match(`len($s) <= 0`).
+		Where(m["s"].Type.Is(`string`)).
+		Report("replace `$$` with `$s == \"\"`")
+	m.Match(`len($s) < 0`).
```
quasilyte, line 176: This is a `badCond` client, isn't it?
len should never return any negative values, so `len(x) < 0` should always be false.

cristaloleg: Yep, you're absolutely correct. Fixed.

quasilyte (approved):

PR #1416, opened 2024-04-29
  checkers/analyzer: remove concurrency from `go/analysis` analyzer

### LINKED ISSUE #1414: check/importShadow: panic due to concurrent access of `PkgObjects` map
Thanks for maintaining these very useful linters 😄 

I've come across a `concurrent map iteration and map write` panic at [`checkers/importShadow_checker.go#L34`](https://github.com/go-critic/go-critic/blob/v0.11.3/checkers/importShadow_checker.go#L34). I'm using the latest version of go-critic (v0.11.3). I think that the `PkgObjects` map is not protected against concurrent accesses.

Unfortunately, I cannot find where the map is also written to concurrently, but should it be protected by a mutex? If this is the case, I'd be happy to contribute a fix.

zakcutner:
There are currently several concurrency issues with the analyzer:

1. `ctx.SetFileInfo` and `c.Check` can be called concurrently. However, go-critic does not support multiple operations on the same context at once.

2. This issue also means that checkers may be called with a context containing the wrong file, because the checker being run in a goroutine may not be complete before `ctx.SetFileInfo` is called for the next file.

3. The same checker is called multiple times in parallel. This is not supported by go-critic, because checkers rely on non-synchronized local state.

It is also worth noting that these issues currently exist even if the `concurrency` flag is set to 1. This is because registering files and running checkers are still executed in parallel, even when concurrency is disabled.

Ultimately, I do not see a good way to integrate concurrency, go-critic and the `go/analysis` framework together, and the current implementation is unsound. Therefore, I have removed the `concurrency` flag and changed both the registration of files and running of checkers to run sequentially.

One potential way to improve this in the future would be to provide a separate `go/analysis` analyzer for each checker, similar to [how staticcheck does](https://pkg.go.dev/honnef.co/go/tools@v0.4.7/staticcheck#pkg-variables). This would allow several go-critic analyzers to be run in parallel using the `go/analysis` framework. However, this would be a much greater breaking change to go-critic's `go/analysis` API, so I have not attempted this for now.

Closes #1414

### DIFF
--- a/checkers/analyzer/analyzer.go
+++ b/checkers/analyzer/analyzer.go
@@ -2,8 +2,6 @@
 package analyzer
 
 import (
-	"runtime"
-
 	"github.com/go-critic/go-critic/linter"
 
 	"golang.org/x/tools/go/analysis"
@@ -23,12 +21,11 @@ var Analyzer = &analysis.Analyzer{
 var DisableCache = false
 
 var (
-	flagGoVersion   string
-	flagEnable      string
-	flagDisable     string
-	flagEnableAll   bool
-	flagDebugInit   bool
-	flagConcurrency int
+	flagGoVersion string
+	flagEnable    string
+	flagDisable   string
+	flagEnableAll bool
+	flagDebugInit bool
 )
 
 var (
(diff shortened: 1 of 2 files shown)


zakcutner:
@quasilyte @cristaloleg Sorry for the ping, I just wanted to ask if either of you might get a chance to take a look at this please?

cristaloleg (approved):
LGTM

cristaloleg:
@quasilyte are you ok with this change?

zakcutner:
@quasilyte @cristaloleg Just bumping this, would it be possible to get this merged?

cristaloleg:
I'm totally fine with it, I will merge it. Thank you for the PR and sorry for the delay 🙌

zakcutner:
No problem, thanks for the help with this!

PR #1434, opened 2024-09-22
checkers/hugeParam: fix: ast.FuncDecl.Type.Results could be nil when checking for String()

Afterous:
Hello,
This will fix a panic when you run go critic with hugeParam and scan a function that is named String, but does not return any value.

The document state the Results could be nil, this change will check for the nil before checking the length.
https://pkg.go.dev/go/ast#FuncType

Please let me know if you need me to write a test or fix anything else.

Thank you again for writing this tool!

example code that when ran with `./gocritic check -enable=hugeParam` will panic
```go
package main

type foo struct {}

func (foo) String() {
  // Empty
}

func main() {
  f := foo{}
  f.String()
}
```

### DIFF
--- a/checkers/hugeParam_checker.go
+++ b/checkers/hugeParam_checker.go
@@ -56,6 +56,7 @@ func (*hugeParamChecker) isImplementStringer(decl *ast.FuncDecl) bool {
 	if decl.Recv != nil &&
 		decl.Name.Name == "String" &&
 		decl.Type != nil &&
+		decl.Type.Params.List != nil && decl.Type.Results.List != nil &&
 		len(decl.Type.Params.List) == 0 &&
 		len(decl.Type.Results.List) == 1 &&
 		astcast.ToIdent(decl.Type.Results.List[0].Type).Name == "string" {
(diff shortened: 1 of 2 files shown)


checkers/hugeParam_checker.go
```diff
@@ -57,14 +57,14 @@ func (*hugeParamChecker) isImplementStringer(decl *ast.FuncDecl) bool {
...
 		len(decl.Type.Params.List) == 0 &&
+		decl.Type.Results != nil &&
 		len(decl.Type.Results.List) == 1 &&
 		astcast.ToIdent(decl.Type.Results.List[0].Type).Name == "string" {
 		return true
 	}
 
 	return false
 }
-
```
cristaloleg, line 67: Please, leave the newline here. Thank you.

Afterous: Sorry about that, the removal of the line has been reverted.

cristaloleg (approved):

Afterous:
@cristaloleg what steps would I need to complete to get this merged into the master branch?

cristaloleg (approved):

PR #1466, opened 2025-03-18
checkers/dupArg: add more patterns

alingse:
I'm implement some [repeat-args](https://github.com/alingse/copyandpaste/blob/main/repeat_args.go#L10-L14) checker, I occurly find it has already implememnt in go-critic's `dupArg`

I add some new functions that should not call with dup args.

btw, I will create another `dupOption` checker PR this week later. see [testcase](https://github.com/alingse/copyandpaste/blob/main/testdata/src/repeatoptions/positive.go#L40-L42)


cristaloleg (approved):
great addition, thanks!

cristaloleg:
Thanks again!

PR #1471, opened 2025-03-21
checkers: add dupOption

alingse:
I first implement it in another copyandpaste linter, i found that the `repeat arg` like `math.Max(a, a)` has already added in go-critic,

so i move the left part `repeat option` here, I have check the open golang github repos, the result can be found here https://github.com/alingse/copyandpaste/issues/7 

of cause, this may report some false-positive case like [docker/compose/blob/main/cmd/compose/compose.go#L399-L407](https://github.com/docker/compose/blob/main/cmd/compose/compose.go#L399-L407)

### DIFF
--- a/checkers/dupOption_checker.go
+++ b/checkers/dupOption_checker.go
@@ -1,0 +1,113 @@
+package checkers
+
+import (
+	"go/ast"
+	"go/token"
+	"go/types"
+
+	"github.com/go-critic/go-critic/checkers/internal/astwalk"
+	"github.com/go-critic/go-critic/linter"
+	"github.com/go-toolsmith/astcast"
+	"github.com/go-toolsmith/astfmt"
+)
+
+func init() {
+	var info linter.CheckerInfo
+	info.Name = "dupOption"
+	info.Tags = []string{linter.DiagnosticTag, linter.ExperimentalTag}
+	info.Summary = "Detects duplicated option func arg"
+	info.Before = `doSomething(name,
+		withWidth(w),
+		withHeight(h),
+		withWidth(w),
+)`
+	info.After = `doSomething(name,
+		withWidth(w),
+		withHeight(h),
+)`
+
+	collection.AddChecker(&info, func(ctx *linter.CheckerContext) (linter.FileWalker, error) {
+		c := &dupOptionChecker{ctx: ctx}
+		return astwalk.WalkerForExpr(c), nil
+	})
+}
+
+type dupOptionChecker struct {
+	astwalk.WalkHandler
+	ctx *linter.CheckerContext
+}
+
+func (c *dupOptionChecker) VisitExpr(expr ast.Expr) {
+	call := astcast.ToCallExpr(expr)
+	variadicArgs, argType := c.getVariadicArgs(call)
+	if len(variadicArgs) == 0 {
+		return
+	}
+
+	if !c.isOptionType(argType) {
+		return
+	}
+
+	dupArgs := c.findDupArgs(variadicArgs)
+	for _, arg := range dupArgs {
+		c.warn(arg)
+	}
+}
+
+func (c *dupOptionChecker) getVariadicArgs(call *ast.CallExpr) ([]ast.Expr, types.Type) {
+	if len(call.Args) == 0 {
+		return nil, nil
+	}
+
+	// skip for someFunc(a, b ...)
+	if call.Ellipsis != token.NoPos {
+		return nil, nil
+	}
+
+	funType := c.ctx.TypesInfo.TypeOf(call.Fun)
+	sign, ok := funType.(*types.Signature)
+	if !ok || !sign.Variadic() {
+		return nil, nil
+	}
+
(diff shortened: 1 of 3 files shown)


checkers/testdata/dupOption/positive_tests.go
```diff
@@ -0,0 +1,62 @@
...
+
+func withHeight(h int) func(c *panel) {
+	return func(c *panel) {
+		c.height = h
+	}
+}
+
+type exportOptions struct{}
+
+type exportOpt func(context.Context, *exportOptions) error
```
cristaloleg, line 38: Can you add a similar case but where `exportOpt` is an alias, literally:

```
type exportOpt = func(context.Context, *exportOptions) error
```

Supporting aliases is often buggy in linters :(

checkers/dupOption_checker.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (c *dupOptionChecker) isOptionType(typeInfo types.Type) bool {
+	typeInfo = typeInfo.Underlying()
+
+	sign, ok := typeInfo.(*types.Signature)
+	if !ok {
+		return false
+	}
+
+	if sign.Params().Len() == 0 {
```
cristaloleg, line 90: Not sure about this one. Looks like the 2nd call should be reported:

```
	_ = newPanel("hello",
		withAutosize(),
		withAutosize(),
	)
```

alingse: it check the type of return result from `withAutoSize()` , not `withAutoSize`'s signature, I add more test case like

`withDefaultWidth()` and `defaultWidth` option, how to create the option is ignored, only check the aftfmt output (code) is same.

checkers/dupOption_checker.go
```diff
@@ -0,0 +1,113 @@
...
+
+	return true
+}
+
+func (c *dupOptionChecker) findDupArgs(args []ast.Expr) []ast.Expr {
+	codeMap := make(map[string]bool)
+	dupArgs := make([]ast.Expr, 0)
+	for _, arg := range args {
+		code := astfmt.Sprint(arg)
+		if !codeMap[code] {
```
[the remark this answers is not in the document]
alingse: done, i start know what you mean? do you mean that
```
if codeMap[code] {
       dupArgs = append(dupArgs, arg)
       continue
}
codeMap[code] = true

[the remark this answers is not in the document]
Comment (cristaloleg) [reply checkers/dupOption_checker.go]: This not gonna work, you need to drop `continue`.

File: checkers/dupOption_checker.go (hunk 2 above)
Comment (cristaloleg) [reply checkers/dupOption_checker.go:90]: Well, `autoSize` can set some field to `true` implicitly. So, param isn't required.

Also, `withDefaultConfig()` which sets bunch of fields at once.

Comment (alingse) [reply checkers/dupOption_checker.go:90]: yes, you are right, if some Option is the type `func() Config`, it can has no params and set some fields, 

but i don't think this is a good `Option`, like this code, `Option1`, `Option2`, `Option3` is not support,  

how do you think? 🤔

```go
type Config struct {
	width  int
	height int
}

type Option1 func()
type Option2 func() error
type Option3 func() (Config, error)
type Option4 func(c *Config)
type Option5 func(c *Config) error
type Option6 func(c Config) Config
```

cristaloleg: I think that's fine for now. Let's see how it will work in the wild.

cristaloleg (approved):
Thanks! I will take another look today in the evening and merge it.

checkers/dupOption_checker.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/go-critic/go-critic/linter"
+	"github.com/go-toolsmith/astcast"
+	"github.com/go-toolsmith/astfmt"
+)
+
+func init() {
+	var info linter.CheckerInfo
+	info.Name = "dupOption"
+	info.Tags = []string{linter.DiagnosticTag, linter.ExperimentalTag}
+	info.Summary = "Detects duplicated option func arg"
```
alexandear, line 18: I propose to extend the checker description.

Something like:
```suggestion
	info.Summary = "Detects duplicated option function arguments in variadic function calls"
```

checkers/dupOption_checker.go
```diff
@@ -0,0 +1,113 @@
...
+			dupArgs = append(dupArgs, arg)
+			continue
+		}
+		codeMap[code] = true
+	}
+	return dupArgs
+}
+
+func (c *dupOptionChecker) warn(arg ast.Node) {
+	c.ctx.Warn(arg, "func arg `%s` is duplicated", arg)
```
alexandear, line 112: I propose to slightly extend the message:
```suggestion
	c.ctx.Warn(arg, "function argument `%s` is duplicated", arg)
```

alexandear: Also, maybe it's makes sense to report the positions of duplicated arguments.

alexandear (approved):

cristaloleg (approved):

PR #1486, opened 2025-04-26
checker/badCond: rhs should be also check for safe expr

### LINKED ISSUE #1485: badCond: reports wrong suspicious condition
Found a bug in `badCond`: 

```
if paramType == lastParam.GetType() && paramType == bgp.BGP_ASPATH_ATTR_TYPE_SEQ {
```

Which indeed looks suspicious but here is the thing: `GetType()` might return `BGP_ASPATH_ATTR_TYPE_SEQ` which is a correct behaviour.

I think in `badCond` we should compare only simple/pure/safe expressions.

codeimmortal:
Added checks for RHS also, in case of unsafe expression we should not throw errors.

@cristaloleg 

Fixes #1485

### DIFF
--- a/checkers/badCond_checker.go
+++ b/checkers/badCond_checker.go
@@ -87,7 +87,8 @@ func (c *badCondChecker) checkExpr(expr ast.Expr) {
 
 func (c *badCondChecker) equalToBoth(lhs, rhs *ast.BinaryExpr) bool {
 	return lhs.Op == token.EQL && rhs.Op == token.EQL &&
-		astequal.Expr(lhs.X, rhs.X)
+		astequal.Expr(lhs.X, rhs.X) &&
+		typep.SideEffectFree(c.ctx.TypesInfo, lhs.Y) && typep.SideEffectFree(c.ctx.TypesInfo, rhs.Y)
 }
 
 func (c *badCondChecker) lessAndGreater(lhs, rhs *ast.BinaryExpr) bool {
--- a/checkers/testdata/badCond/negative_tests.go
+++ b/checkers/testdata/badCond/negative_tests.go
@@ -2,6 +2,8 @@ package checker_test
 
 func getIntPtr(v *int) {}
 
+func newError() error { return nil }
+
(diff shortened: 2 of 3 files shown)


checkers/badCond_checker.go
```diff
@@ -86,8 +87,20 @@ func (c *badCondChecker) checkExpr(expr ast.Expr) {
 }
 
 func (c *badCondChecker) equalToBoth(lhs, rhs *ast.BinaryExpr) bool {
-	return lhs.Op == token.EQL && rhs.Op == token.EQL &&
-		astequal.Expr(lhs.X, rhs.X)
```
cristaloleg, line 90: Hm, feels like 
```
return lhs.Op == token.EQL && rhs.Op == token.EQL &&
		astequal.Expr(lhs.X, rhs.X) &&
        typep.SideEffectFree(info, rhs.X)
```

is what we need, isn't it?

codeimmortal: Yes this approach will also solve it. Shall we check both Y value for both sideEffectFree. That will be perfect. I think typep.SideEffectFree(info, rhs.X) was mistake, you want rhs.Y to be validated.

func (c *badCondChecker) equalToBoth(lhs, rhs *ast.BinaryExpr) bool {
	return lhs.Op == token.EQL && rhs.Op == token.EQL &&
		astequal.Expr(lhs.X, rhs.X) &&
		typep.SideEffectFree(c.ctx.TypesInfo, lhs.Y) && typep.SideEffectFree(c.ctx.TypesInfo, rhs.Y)
}

cristaloleg: Yeah, I might be wrong with the real expression but looks like everything can be done in a few additional lines if I'm not mistaken.

codeimmortal: Added condition. Thanks for suggestion. Now it is clean.

quasilyte (approved):

cristaloleg (approved):

PR #1488, opened 2025-05-11
Add zeroByteRepeat for avoiding bytes.Repeat([]byte{0}, x)

codeimmortal:
make already returns a zeroed slice, no need to use bytes.Repeat for that. Should be easily achievable via ruleguard.

Also one more condition if value is const 0 inside bytes.Repeat([]byte{<const0>}, x), that also should cover in rule.

Let me know if second condition not needed.


cristaloleg:
Test failure https://github.com/go-critic/go-critic/actions/runs/14953260890/job/42005486293?pr=1488#step:6:678

codeimmortal:
Added missed value in test case (TestStableList).

cristaloleg:
Thanks!

PR #1491, opened 2025-06-01
docs/readme: add Homebrew install instructions

krehel:
Hi! 👋 

I noticed there weren't instructions for installing via Homebrew on macOS, so adding them.

I also adjusted some spacing in a couple places where the Markdown linter recommended to do so.

Thank you for such a useful tool!


cristaloleg (approved):
Great, thanks!

PR #1495, opened 2025-07-01
checkers: add suspiciousIota (#1469)

### LINKED ISSUE #1469: Report suspicious iota usages
This code may seem OK, but it contains an error:
```go
const (
	maxCommandID = 9999
	
	ValueGrayFlag = iota + (maxCommandID + 1)
	ValueBlueFlag
	ValueTealFlag
)
```
`ValueGrayFlag` will be 10001 instead of the implied 10000.

In order to fix it, the unrelated const needs to be moved away (or +1 should be removed, but it's sloppy).
```go
const maxCommandID = 9999

const (
	ValueGrayFlag = iota + (maxCommandID + 1)
	ValueBlueFlag
	ValueTealFlag
)
```

Therefore, we can assume that this is a bug pattern. There are legitimate usages of intermixed iota and other constants, but I would predict that we should at least try to find the false positive/negative ratio. Enabling this checker as an experimental addition to go-critic should be good enough.

This linter should report any constants that appear in `const ()` spec where `iota` is also used. All preceding iota constants should be reported (there should be 0).

We should be careful to emphasize on a fact that moving the const away is not always a full solution - iota constants should be revisited (their values might change).

atlet99:
## Add suspiciousIota checker

Detects misleading iota usage where constants before `iota` cause unexpected values.

```go
const (
    maxID = 9999
    Flag = iota + 10000  // Bug: equals 10001, not 10000 (iota=1 here)
)
```

Also catches redundant `= iota` assignments and mixed iota/explicit patterns.

Research shows 50% of devs get this wrong. Tagged experimental.

Fixes #1469

### DIFF
--- a/checkers/suspiciousIota_checker.go
+++ b/checkers/suspiciousIota_checker.go
@@ -1,0 +1,306 @@
+package checkers
+
+import (
+	"go/ast"
+	"go/constant"
+	"go/token"
+	"strings"
+
+	"github.com/go-critic/go-critic/checkers/internal/astwalk"
+	"github.com/go-critic/go-critic/linter"
+
+	"github.com/go-toolsmith/astfmt"
+)
+
+func init() {
+	var info linter.CheckerInfo
+	info.Name = "suspiciousIota"
+	info.Tags = []string{linter.StyleTag, linter.ExperimentalTag}
+	info.Summary = "Detects suspicious iota usage patterns"
+	info.Before = `
+const (
+	A = iota
+	B = iota // redundant iota
+	C = iota // redundant iota
+)`
+	info.After = `
+const (
+	A = iota
+	B        // auto-increment
+	C        // auto-increment
+)`
+
+	collection.AddChecker(&info, func(ctx *linter.CheckerContext) (linter.FileWalker, error) {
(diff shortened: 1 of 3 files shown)


atlet99:
> Thanks @atlet99
> 
> I feel like anyone reviewing the PR should read the discussion that happened in the issue #1469 or on Slack discussion that happened in the following thread.
> 
> https://gophers.slack.com/archives/C029RQSEE/p1742479292573629?thread_ts=1742479292.573629&cid=C029RQSEE
> 
> I didn't review yet, I'm just raising the point it should be reviewed carefully.
> 
> But yes, I agree most devs are using it wrong. I was too until I got it.
> 
> So I'm happy someone worked on providing a linter

Thank you! I’m really glad I could contribute and help improve things. I've carefully read through the discussion in issue #1495 and took it into account while working on the PR. Unfortunately, I missed the Slack thread... that's my oversight, and I appreciate you pointing it out. Totally agree - it's easy to misuse this, and I'm happy to help by introducing a linter that can guide devs in the right direction.

ccoVeille:
> Thank you for the PR, i will take a look today-tomorrow.

@cristaloleg could you review it then?

PR #1522, opened 2025-12-29
perf: cache ruleguard engines for embedded rules

dgageot:
Previously, each call to NewChecker for an embedded ruleguard rule would create a new ruleguard.Engine and call LoadFromIR, which performs expensive type imports via go/internal/srcimporter. When golangci-lint runs go-critic on a large codebase, this happens repeatedly for every package being analyzed, causing significant performance degradation.

This change introduces a cachedEngine struct that lazily initializes each rule group's engine on first use (via sync.Once) and reuses it for all subsequent checker instances. The engine itself is stateless with respect to the files being analyzed - only the RunContext passed to WalkFile contains package-specific information.

Benchmark results on a medium-sized Go project (~100 packages):
- Before: 24.46s wall clock, 216.90s total CPU time
- After:  6.74s wall clock, ~59s total CPU time
- Improvement: 72% reduction in wall clock time

The bottleneck was identified via CPU profiling, which showed findTypeNoCache and srcimporter operations consuming ~50% of CPU time due to repeated engine initialization.

Assisted-By: cagent

### DIFF
--- a/checkers/embedded_rules.go
+++ b/checkers/embedded_rules.go
@@ -6,6 +6,7 @@ import (
 	"go/build"
 	"go/token"
 	"os"
+	"sync"
 
 	"github.com/go-critic/go-critic/checkers/rulesdata"
 	"github.com/go-critic/go-critic/linter"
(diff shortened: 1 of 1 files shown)


cristaloleg:
Hey, thanks for the PR, looks reasonable. Did you run it on a "dirty" codebase (with the reported issues) and seen the same output as for the released `go-critic` version?

dgageot:
@cristaloleg I tested it only through `golangci-lint`. Each time I made sure the cache was empty. The numbers I shared are between the released `go-critic` and the patched `go-critic`.

However, since I ran this with a lot of AI assistance, that would be comforting if other people can reproduce my results :-)

dgageot:
@cristaloleg anything I can do to help get this merged?

cristaloleg:
Merged, thank you for the PR and sorry for the delay.

PR #1527, opened 2026-04-05
feat: Add GOOS linter for issue 1472

Afterous:
This will add a linter to catch a comparison for a GOOS which doesn't exist in checkers/goosChecker_syslist_gen.go.

 goosChecker_syslist_gen.go is generated by go generate and uses the version of the toolchain installed.


checkers/goosChecker_gen/gen.go
```diff
@@ -0,0 +1,138 @@
...
+		return fmt.Errorf("read syslist.go: %v", err)
+	}
+
+	fset := token.NewFileSet()
+	f, err := parser.ParseFile(fset, syslistPath, src, 0)
+	if err != nil {
+		return fmt.Errorf("parse syslist.go: %v", err)
+	}
+
+	goos, err := extractMapKeys(f, "KnownOS")
```
Afterous, line 64: The assumption is the variable won't change.

However anything under the `internal/` can change and break this linter.

Would we rather just handle a hard coded list instead of code generation?

Afterous:
Ci failures is due to oldstable seeing a different generated file I think. 

We could remove the version in the code generation, but if we do that if the GOOS changes between releases we will return another error if the support GOOS changes (as stable will differ than oldstable, unless it is back ported by the go team).

How do you think this should be addressed? I'm leaning on just leveraging a hardcoded list that doesn't rely on code generation.

PR #1529, opened 2026-06-14
checkers: extend sprintfQuotedString with %#q case

alexandear:
This PR proposes extending `sprintfQuotedString` with the `%#q` case:

Before:
```go
fmt.Printf("`%s`", "foo")
```

After:
```go
fmt.Printf("%#q", "foo")
```

From https://pkg.go.dev/fmt:

> // Strings are formatted with %v and %s as-is, with %q as quoted strings,
> // and %#q as backquoted strings.
> placeholders := `foo "bar"`
> fmt.Printf("%v %s %q %#q\n", placeholders, placeholders, placeholders, placeholders)
> // Result: foo "bar" foo "bar" "foo \"bar\"" `foo "bar"`


cristaloleg (approved):
Thanks!

Issue #1245, opened 2022-08-08
typeUnparen: reports false-positive `(func())(nil)`


tdakkota:
`typeUnparen` reports `could simplify (func()) to func()`, but `func() (nil)` is invalid function literal and definitely is not the same expression.

Reproducer:
```go
package yaml

func _() {
	_ = (func())(nil)
}
```

Discovered [here](https://github.com/tdakkota/yamlx/commit/8b278946d3a6619f6493e1a7423f5d3f05dcc9b8#diff-5afd41e7e0a70c3a6f0e1572f6938d9aacf5bc590033f7e798a0fb6f741e51ddR42).

`go-critic` version (I've installed it using `go install`, so there is no `gocritic version` information):
```
$ go version -m (where.exe gocritic)
C:\Users\tdakkota\go\bin\gocritic.exe: go1.19
        mod     github.com/go-critic/go-critic  v0.6.3  h1:abibh5XYBTASawfTQ0rA7dVtQT+6KzpGqb/J+DxRDaw=
        ...
```

The `golangci-lint` version has the same issue.

Issue #1035, opened 2021-04-06
Suggesting typeUnparen when not equivalent


bruno-anjos:
I stumbled upon this suggestion:

```
typeUnparen: could simplify (<-chan *WebsocketMsg) to <-chan *WebsocketMsg (gocritic)
```

for this line:

```
messagesQueue := make(chan (<-chan *WebsocketMsg), 10)
```

From what i saw in the go documentation this is not right since the arrow is applied to the leftmost type
([see ](https://golang.org/ref/spec#Channel_types))

There it says the following

```
The <- operator associates with the leftmost chan possible:

chan<- chan int    // same as chan<- (chan int)
chan<- <-chan int  // same as chan<- (<-chan int)
<-chan <-chan int  // same as <-chan (<-chan int)
chan (<-chan int)
```

And therefore if i remove the parenthesis i would go from a channel of receive-only channels to a send-only channel of channels. Can you please confirm i might be missing something.

quasilyte:
Interesting case!
Thank you for reporting this.
I'll try to roll a solution in the near time.
(If nobody beats me to it.)

PR #1536, opened 2026-08-08
checkers/internal/astwalk: keep the parenthesis of a receive-only channel type

### LINKED ISSUE #1428: typeUnparen false positive for `<- chan`
Looks to be similar cause as #1035 but a slightly different case.
Tested on `v0.11.5-0.20240722075729-97e2bc38fa41`.

`typeUnparen` suggests removing parentheses around `<-chan` types that would actually break the code, e.g.
```go
package main

func main() {
	a := (<-chan int)(nil)
	recv(a)
	b := (chan<- int)(nil)
	send(b)
}

func recv(c <-chan int) {
	<-c
}

func send(c chan<- int) {
	c <- 0
}
```
```console
$ gocritic check -enableAll
./main.go:4:7: typeUnparen: could simplify (<-chan int) to <-chan int
./main.go:6:7: typeUnparen: could simplify (chan<- int) to chan<- int
$ 
```

Following the suggest changes, we get a compiler error because `a := <-chan int(nil)` is now assigning the result of `<-nil`, an int value, to `a`.

```console
$ go build main.go
# command-line-arguments
./main.go:5:7: cannot use a (variable of type int) as <-chan int value in argument to recv
```

I guess `chan<- int` is less ambiguous so isn't broken the same way, although it's arguable that keeping parentheses there would be more readable...

Eljees:
Fixes #1428.

## The report

`typeUnparen` suggests simplifying `(<-chan int)(nil)` to `<-chan int(nil)`. Applying that
suggestion does not compile.

Checked with the compiler rather than assumed:

| source | result |
|---|---|
| `a := (<-chan int)(nil)` | compiles |
| `a := <-chan int(nil)` (the suggested edit) | **fails**: `cannot use a (variable of type int) as <-chan int value in argument to recv` |
| `b := chan<- int(nil)` | compiles |

So the parenthesis is required for the receive direction and redundant for the send direction —
which matches the spec: *"If the type starts with the operator `*` or `<-`, or if the type
starts with the keyword `func` and has no result list, it must be parenthesized when necessary
to avoid ambiguity."*

## Cause

`typeExprWalker.inspectInner` is what skips the outermost parenthesis of a conversion, and its
comment already says so. Its whitelist implements only two of the three spec cases:

```go
	shouldInspect := ok &&
		typep.IsTypeExpr(w.info, parens.X) &&
		(astp.IsStarExpr(parens.X) || astp.IsFuncType(parens.X))
```

`*` (`(*int)(nil)`) and `func` (`(func())(nil)`, #1245) are handled; `<-` is missing. For
`(<-chan int)(nil)` the walker therefore descends normally, reaches the `*ast.ParenExpr` case,
and `typeUnparen` strips a parenthesis the code cannot compile without.

`inspectInner` is only reached from the `*ast.CallExpr` and `*ast.SelectorExpr` cases —
conversions and method expressions — so this does not affect places where the parenthesis
really is redundant, e.g. `var _ (<-chan int)`.

Note the `*ast.ChanType` branch inside `typeUnparen_checker.go` (the fix for #1035, nested
channels such as `chan (<-chan T)`) is a different code path and keeps working; its fixture
`channelIssue1035` stays green.

## Fix

Add the third spec case to the whitelist, restricted to `ast.RECV` because the compiler run
above shows `chan<-` needs no parenthesis.

## Tests

- `checkers/testdata/typeUnparen/negative_tests.go` — `convIssue1428`, the reported snippet.
  Before the fix: `testdata/typeUnparen/negative_tests.go:105: unexpected warn: could simplify
  (<-chan int) to <-chan int`.
- `checkers/testdata/typeUnparen/positive_tests.go` — `convIssue1428SendOnly`, asserting that
  `(chan<- int)(nil)` is **still** reported, so the fix cannot silently widen to both
  directions. (This one earned its place: an earlier draft with the expectation marker on the
  wrong line failed with `unexpected warn` at that exact spot, which is also the proof the
  warning is really emitted there.)

Verified on `325d070` with Go 1.26:

- `go test -count=1 -run=/typeUnparen ./checkers` — red before, `ok` after;
- `go test -race -count=1 ./...` — all packages `ok`, so `ptrCast`, `funcIssue1245`,
  `channelIssue1035` and the rest of the golden fixtures are unaffected;
- `go generate ./...` leaves the tree clean;
- `go-critic check -enableAll ./...` on the patched tree — no findings;
- `gofmt` clean for the touched files.

---

Prepared with AI assistance. Every claim above was verified by running it.

### DIFF
--- a/checkers/internal/astwalk/type_expr_walker.go
+++ b/checkers/internal/astwalk/type_expr_walker.go
@@ -95,11 +95,22 @@ func (w *typeExprWalker) walk(x ast.Node) bool {
 	return true
 }
 
+// isRecvOnlyChanType reports whether x is a receive-only channel type, `<-chan T`.
+//
+// The Go spec requires a parenthesis around a conversion type that starts with
+// `*`, `<-`, or the `func` keyword: without it `(<-chan int)(nil)` parses as
+// `<-(chan int(nil))` and does not compile. A send-only `chan<- int(nil)` is
+// unambiguous, so only the receive direction is special here.
+func isRecvOnlyChanType(x ast.Expr) bool {
+	ch, ok := x.(*ast.ChanType)
+	return ok && ch.Dir == ast.RECV
+}
+
 func (w *typeExprWalker) inspectInner(x ast.Expr) bool {
 	parens, ok := x.(*ast.ParenExpr)
 	shouldInspect := ok &&
 		typep.IsTypeExpr(w.info, parens.X) &&
-		(astp.IsStarExpr(parens.X) || astp.IsFuncType(parens.X))
+		(astp.IsStarExpr(parens.X) || astp.IsFuncType(parens.X) || isRecvOnlyChanType(parens.X))
 	if shouldInspect {
 		ast.Inspect(parens.X, w.walk)
 		return false
--- a/checkers/testdata/typeUnparen/negative_tests.go
+++ b/checkers/testdata/typeUnparen/negative_tests.go
@@ -100,3 +100,7 @@ func channelIssue1035() {
 func funcIssue1245() {
 	_ = (func())(nil)
 }
+
+func convIssue1428() {
+	_ = (<-chan int)(nil)
+}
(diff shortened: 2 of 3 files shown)


cristaloleg (approved):
SGTM!

Eljees:
@cristaloleg thanks for the review. This one and #1535 have both been approved and mergeable since 9 August with nothing since - is there anything else you would like from me, or are they waiting on a release window? Happy to rebase either if they have gone stale in the meantime.

